PR #2, opened 2016-04-25
Reduce dependencies, enable systems with fewer available commands

chsterz:
Systems using sh only now can execute the heaptrack startscript.
Cat-to-procfs should be available on Linux systems that do not provide the ps command.

### DIFF
--- a/heaptrack.sh.cmake
+++ b/heaptrack.sh.cmake
@@ -59,18 +59,15 @@ client=
 
 while true; do
     case "$1" in
-        "-d") ;&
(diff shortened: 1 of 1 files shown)


milianw:
We don't accept pull requests via GitHub in KDE. Please close this here and either send me the patches via email, or better yet use phabricator.

milianw:
can you close this review? I incorporated your changes in the meantime

thanks

chsterz:
done, thanks

PR #3, opened 2016-08-24
Changes required to fix compilation errors under DevToolset4 on CentO…

pianofab:
…S, gcc 5.2.1

https://www.softwarecollections.org/en/scls/rhscl/devtoolset-4/


tracetree.h
```diff
@@ -76,7 +76,7 @@ class TraceTree
             if (it == parent->children.end() || it->instructionPointer != ip) {
                 index = m_index++;
                 it = parent->children.insert(it, {ip, index, {}});
-                fprintf(out, "t %" PRIxPTR " %x\n", reinterpret_cast<uintptr_t>(ip), parent->index);
```
milianw: this seems wrong to me

pianofab: Please explain what seems wrong. Reading material: http://stackoverflow.com/questions/1255099/whats-the-proper-use-of-printf-to-display-pointers-padded-with-0s

milianw: %p is printed as 0xABCDEF but we only want ABCDEF without the leading 0x. PRIxPTR is what we want, and that is also the accepted answer of the stackoverflow question you are linking to?

pianofab: Now I understand what you are saying, but the "accepted answer" on StackOverflow is not always the correct answer -- and also, sometimes the correct answer may vary depending on the particular requirements. If you want maximum portability, the correct answer is actually slightly below (quoted at bottom here). I should mention that at the time I was probably trying something which I later realized was not part of your scope (compiling on a 64-bit OS). I simply saved this change for some other team in my company to pick up, but then our investigation moved into a different direction. I am currently a bit busy but if it helps I can try to build heaptrack again to see what the exact problem is.
>>>

The only portable way to print pointer values is to use the "%p" specifier, which produces implementation-defined output. (Converting to uintptr_t and using PRIxPTR is likely to work, but (a)  uintptr_t was introduced in C99, so some compilers may not support it, and (b) it's not guaranteed to exist even in C99 (there might not be an integer type big enough to hold all possible pointer values.

So use "%p" with sprintf() (or snprintf(), if you have it) to print to a string, and then print the string padding with leading blanks.

milianw: Please note that I'm depending on C++11, and I'm perfectly fine with depending on C99. If a platform does not support PRIxPTR then heaptrack will probably fail elsewhere:

- I'm using uintptr_t extensively
- I'm even assuming that uintptr_t will always fit into a uint64_t when analyzing the data

So, again: If you want me to merge your pull request, please take out this hunk. Adding the cinttypes include is a good change by itself and I would like to have this in, now. Otherwise I'll do it myself.

Thanks a lot

milianw:
Can you please squash these changes into one?

PR #7, opened 2017-03-03
Update README.md

muojp:
This `s` matters when specifying `--with-libraries` option on boost's bootstrap.sh.

I ran into this when I was trying to build boost-1.41.0 required by heaptrack on Ubuntu 16.04.


milianw:
Cherry-picked into 1.0, then merged into master from there. Thanks!

PR #8, opened 2017-04-05
Add missing include directory.

pokowaka:
- Boost headers are at $BOOST_ROOT which might not be on the default
  include path.


milianw:
Thanks, applied!

PR #11, opened 2018-02-08
Fix crash if an ELF module has no symbol table

aejsmith:
Set libbacktrace's dummy callback function to avoid it calling a null pointer.


milianw:
Thanks, merged!

PR #14, opened 2018-12-10
Encoding of values in /proc/self/statm is actually decimal

sebastianbiallas:


milianw:
WTF, how did I manage to screw that one up :D

Thanks a lot Sebastian, I'll apply this asap. Cheers

milianw:
merged, thanks a lot

PR #17, opened 2018-12-18
Build GUI for non-Linux systems

blakharaz:
Change the cmake files so that the GUI and print part of heaptrack can be built on a non-Linux system, e.g. on macOS. On non-Linux systems everything but heaptrack_print and heaptrack_gui is skipped, also the tests.
Add a short description for macOS using homebrew.

### DIFF
--- a/3rdparty/CMakeLists.txt
+++ b/3rdparty/CMakeLists.txt
@@ -2,8 +2,10 @@ if (ECM_FOUND)
     include(ECMEnableSanitizers)
 endif()
 
-add_subdirectory(libbacktrace)
+if(HEAPTRACK_BUILD_TRACK)
+    add_subdirectory(libbacktrace)
+endif()
 
 if (ZSTD_FOUND)
-add_subdirectory(boost-zstd)
+    add_subdirectory(boost-zstd)
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,30 +23,44 @@ set(HEAPTRACK_FILE_FORMAT_VERSION 2)
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 
 include(FeatureSummary)
-find_package(Libunwind REQUIRED)
-find_package(Boost 1.41.0 COMPONENTS system filesystem) # option first
-find_package(Boost 1.41.0 REQUIRED COMPONENTS iostreams program_options) # then required
+find_package(Boost 1.41.0 COMPONENTS system filesystem iostreams)
 find_package(Threads REQUIRED)
 find_package(ZLIB REQUIRED)
-find_package(Zstd)
+
+if (${BOOST_iostreams_FOUND})
+    find_package(Zstd)
+endif()
 set_package_properties(Zstd PROPERTIES TYPE RECOMMENDED PURPOSE "Zstandard offers better (de)compression performance compared with gzip/zlib, making heaptrack faster and datafiles smaller.")
+
+if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
+    set(HEAPTRACK_BUILD_TRACK_DEFAULT ON)
+else()
+    set(HEAPTRACK_BUILD_TRACK_DEFAULT OFF)
+endif()
+
+option(
+  HEAPTRACK_BUILD_TRACK
+  "Disable this option to skip building the tracker part for heaptrack, e.g. to only build the GUI."
+  ${HEAPTRACK_BUILD_TRACK_DEFAULT}
+)
+
+if (CMAKE_CROSSCOMPILING)
+    set(HEAPTRACK_BUILD_ANALYZE_DEFAULT OFF)
+else()
+    set(HEAPTRACK_BUILD_ANALYZE_DEFAULT ON)
+endif()
+
+option(
+  HEAPTRACK_BUILD_PRINT
+  "Disable this option to skip building heaptrack_print, e.g. when you're cross-compiling."
+  ${HEAPTRACK_BUILD_ANALYZE_DEFAULT}
+)
 
 option(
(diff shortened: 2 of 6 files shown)


3rdparty/CMakeLists.txt
```diff
@@ -1,8 +1,10 @@
-if (ECM_FOUND)
-    include(ECMEnableSanitizers)
-endif()
+if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
+    if (ECM_FOUND)
```
milianw: the ECM_FOUND conditional should be outside the branch, such that we can still build with sanitizers on non-linux platforms

3rdparty/CMakeLists.txt
```diff
@@ -1,8 +1,10 @@
-if (ECM_FOUND)
-    include(ECMEnableSanitizers)
-endif()
+if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
```
milianw: Please introduce a variable for this, instead of repeating the check multiple times. Something like BUILD_TRACKER

CMakeLists.txt
```diff
@@ -96,6 +84,22 @@ file(RELATIVE_PATH LIB_REL_PATH
 
 set(ECM_ENABLE_SANITIZERS "" CACHE STRING "semicolon-separated list of sanitizers to enable for code that is not injected into client applications")
 
+if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
```
milianw: introduce BUILD_TRACKER here

src/analyze/CMakeLists.txt
```diff
@@ -1,5 +1,7 @@
-if (ECM_FOUND)
-    include(ECMEnableSanitizers)
+if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
```
milianw: dito above regarding sanitizers

3rdparty/CMakeLists.txt
blakharaz: For some reason that didn't work with the extra-cmake-modules coming from homebrew. I'll investigate what is the problem there. I agree sanitizers should work for macOS as well.

src/analyze/CMakeLists.txt
blakharaz: done

blakharaz:
On top of your comments I made the change a bit larger and added more configuration flags in the latest pushed version. It's now possible to skip building the tests, only build the tracker, or only build the analyzer.

CMakeLists.txt
```diff
@@ -23,31 +23,51 @@ set(HEAPTRACK_FILE_FORMAT_VERSION 2)
...
 option(
-  HEAPTRACK_BUILD_GUI
-  "Disable this option to skip building the Qt5 / KF5 based GUI for heaptrack."
+  HEAPTRACK_BUILD_TRACK
+  "Disable this option to skip building the tracker part for heaptrack, e.g. to only build the GUI."
+  ${HEAPTRACK_BUILD_TRACK_DEFAULT}
+)
+
+option(
+  HEAPTRACK_BUILD_TESTS
```
milianw: shouldn't be needed, afaik there's BUILD_TESTING which is defaulted to on. You can specify that to off if needed already, no?

src/analyze/CMakeLists.txt
```diff
@@ -21,13 +23,24 @@ target_link_libraries(sharedprint LINK_PUBLIC
 )
 
 if (ZSTD_FOUND)
-target_link_libraries(sharedprint LINK_PUBLIC
-    boost-zstd
+    target_link_libraries(sharedprint LINK_PUBLIC
+        boost-zstd
 )
```
milianw: if you indent the above, then also indent the line here please

CMakeLists.txt
blakharaz: That's right. I changed it

src/analyze/CMakeLists.txt
blakharaz: Done.

README.md
```diff
@@ -107,6 +107,36 @@ of the CMake command, as it will tell you about missing dependencies!
...
+On macOS the dependencies can be installed easily using [homebrew](http://brew.sh) and the [KDE homebrew tap](https://github.com/KDE-mac/homebrew-kde).
+
+    # prepare tap
+    brew tap kde-mac/kde
+    "$(brew --repo)/Library/Taps/kde-mac/homebrew-kde/tools/all-fixes.sh"
+    
+    # install dependencies
+    brew install kde-mac/kde/kf5-extra-cmake-modules kde-mac/kde/kf5-kcoreaddons kde-mac/kde/kf5-ki18n \
+                 kde-mac/kde/kf5-kitemmodels kde-mac/kde/kf5-threadweaver kde-mac/kde/kf5-kconfigwidgets \
+                 kde-mac/kde/kf5-kio kde-mac/kde/kf5-kdiagram \
```
ShiKaiWi: It seems `kde-mac/kde/kf5-kdiagram` is not found and `kde-mac/kde/kdigram` works for me.

milianw:
Thanks, I squashed the fixup and commit it all now!

YangZhou1997:
I tried to install heaptrack on macOS Mojave 10.14.4. Everything seems going well, but both 
```CMAKE_PREFIX_PATH=/usr/local/opt/qt PATH=$PATH:/usr/local/opt/gettext/bin cmake ..``` and 
```cmake -DCMAKE_BUILD_TYPE=Release ..``` show 
```
-- The following OPTIONAL packages have not been found:

 * KF5KIO
 * KF5
```

That causes being unable to ```make heaptrack_gui``` since gui relies on the two packages. 

Not sure how to solve it; any suggestions are highly appreciated!

milianw:
install the kde frameworks 5 packages from homebrew or similar (afaik, I'm not a mac person)

PR #18, opened 2019-05-25
Added --output option

jgaa:
I added an option to specify the file-name for the data-file.

I need that in order to use heaptrack in docker/kubernetes deployments, where the pid is always the same, and the output directory is shared among containers.

### DIFF
--- a/src/track/heaptrack.sh.cmake
+++ b/src/track/heaptrack.sh.cmake
@@ -96,9 +96,13 @@ while true; do
             exit 0
             ;;
         "-o" | "--output-file")
-            output=$2
-            if [ -z "$output" ]; then
+            if [ -z "$2" ]; then
                 echo "Missing output argument."
+                exit 1
+            fi
+            output=$(readlink -f $2)
+            if [ -d "$output" ]; then
+                echo "Please specify a file-name or a full path-name for output."
                 exit 1
             fi
             shift 2
--- a/src/track/heaptrack.sh.cmake
+++ b/src/track/heaptrack.sh.cmake
@@ -52,6 +52,7 @@ usage() {
     echo "                 to the debuggee."
     echo "  -h, --help     Show this help message and exit."
     echo "  -v, --version  Displays version information."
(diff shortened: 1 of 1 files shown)


src/track/heaptrack.sh.cmake
```diff
@@ -94,6 +95,14 @@ while true; do
             usage
             exit 0
             ;;
+        "-o" | "--output-file")
+            output=$2
+            if [ -z "$output" ]; then
+                echo "Missing output argument."
+                exit 1
+            fi
+            shift 2
```
milianw: We will need to turn the potentially-relative-path into an absolute one, using `readlink -f`. Otherwise, when you inject into an already running application, a relative path will be resolved against the CWD of the application which is highly unexpected to the user. Can you fix that please?

src/track/heaptrack.sh.cmake
```diff
@@ -96,11 +96,15 @@ while true; do
...
             ;;
         "-o" | "--output-file")
-            output=$2
-            if [ -z "$output" ]; then
+            if [ -z "$2" ]; then
                 echo "Missing output argument."
                 exit 1
             fi
+            output=$(readlink -f $2)
+            if [ -d "$output" ]; then
```
milianw: will this catch the case where `$2` points to an invalid path (e.g. a subdir that doesn't exist) and then `output` 
becomes empty? Or do we need to check for ` || [ -z "$output" ]` too?

milianw:
thanks, pushed!

src/track/heaptrack.sh.cmake
jgaa: It falls back to the default output path.

src/track/heaptrack.sh.cmake
```diff
@@ -94,6 +95,18 @@ while true; do
             usage
             exit 0
             ;;
+        "-o" | "--output-file")
```
danilaml, line 98: Noticed that it checks for `output-file` while help uses `output`.

PR #19, opened 2019-05-26
Checks if leaked memory resident in RAM.

tomaszduda23:
Previously all virtual memory that is not released is considered as a leak.
This is a problem in most of the cases but not all.
Sometimes we would like to see only physical memory (RAM) which leaks. E.g.:
int* a = new int[1024*1024*1024];
int* b = new int[1024];
for(size_t i = 0; i < 1024; ++i)
    b[1024] = 0xdeadbeef;
"b" is more worrying than "a" since "a" does not leak RAM (at least on Linux by defaul).
"b" - requires one or more pages of RAM. When "a" does not use RAM at all (https://marek.vavrusa.com/memory/).
If you care more about RAM than virtual memory you should take a look.

The change introduce category "In RAM" which shows memory that leaks and is present in RAM.

### DIFF
--- a/src/track/libheaptrack.cpp
+++ b/src/track/libheaptrack.cpp
@@ -24,6 +24,7 @@
 
 #include "libheaptrack.h"
 
+#include <alloca.h>
 #include <cstdio>
 #include <cstdlib>
 #include <fcntl.h>
@@ -32,7 +33,9 @@
 #include <signal.h>
 #include <stdio_ext.h>
 #include <sys/file.h>
+#include <sys/mman.h>
 #include <syscall.h>
+#include <unistd.h>
 
 #include <atomic>
 #include <cinttypes>
@@ -276,6 +279,7 @@ public:
             pthread_atfork(&prepare_fork, &parent_fork, &child_fork);
 
             atexit([]() {
+                // TODO: it does not work
                 if (s_forceCleanup) {
                     return;
                 }
@@ -332,6 +336,7 @@ public:
 
         writeTimestamp();
         writeRSS();
+        writeInRam();
 
         s_data->out.flush();
         s_data->out.close();
@@ -398,6 +403,103 @@ public:
         //       the RSS numbers with heaptrack-internal data
 
         s_data->out.writeHexLine('R', rss);
+    }
+
+    int saveInRam(size_t address, size_t size, size_t pageSize)
+    {
+        if (0 == size) {
+            return 0;
+        }
+
+        debugLog<VeryVerboseOutput>("save in RAM %zx %zx", address / pageSize, size / pageSize);
+        if (!s_data->out.write(" %zx %zx", address / pageSize, size / pageSize)) {
+            return 1;
+        }
+
+        return 0;
+    }
+
+    void writeInRam()
+    {
+        if (!s_data || !s_data->out.canWrite() || s_data->procMaps == -1) {
+            return;
+        }
+
+        lseek(s_data->procMaps, 0, SEEK_SET);
+
+        const int BUF_SIZE = PATH_MAX + 1024;
+        char line[BUF_SIZE + 1];
+        size_t len = 0;
+
+        size_t pageSize = sysconf(_SC_PAGESIZE);
+
+        if (!s_data->out.write("P")) {
+            return;
+        }
+
+        while (true) {
+            switch (read(s_data->procMaps, &line[len], 1)) {
+            case -1:
+                fprintf(stderr, "WARNING: Failed to read value from /proc/self/maps, %s.\n", strerror(errno));
+                close(s_data->procMaps);
+                s_data->procMaps = -1;
+                return;
+            case 0:
+                s_data->out.write("\n");
+                return;
+            }
+            if (line[len] != '\n') {
+                ++len;
+                continue;
+            }
+            line[len] = 0;
+            len = 0;
+
+            size_t from, to, pgoff;
+            unsigned int major, minor;
+            unsigned long ino;
+            char path[PATH_MAX];
+            char flags[4];
+            if (sscanf(line, "%zx-%zx %4c %zx %x:%x %lu %s", &from, &to, flags, &pgoff, &major, &minor, &ino, path)
+                < 7) {
+                fprintf(stderr, "WARNING: Failed to parse value from /proc/self/maps '%s'.\n", line);
+                close(s_data->procMaps);
+                s_data->procMaps = -1;
+                return;
+            }
+            size_t length = to - from;
+            size_t count = (length + pageSize - 1) / pageSize;
+            unsigned char* vec = static_cast<unsigned char*>(alloca(count));
+            if (mincore(reinterpret_cast<void*>(from), length, vec) == -1) {
+                if (errno == ENOMEM && strcmp(path, "[vsyscall]") == 0) {
+                    // For some reason mincore returns ENOMEM for [vsyscall].
+                    // It does not matter in our case.
+                    continue;
+                }
+                fprintf(stderr, "WARNING: Failed to mincore at %zx, length %zu, name %s, %s.\n", from, length, path,
+                        strerror(errno));
+                close(s_data->procMaps);
+                s_data->procMaps = -1;
+                return;
+            }
+
+            size_t address = from;
+            size_t size = 0;
+            for (size_t i = 0; i < count; ++i) {
+                if ((vec[i] & 1) == 0) {
+                    if (saveInRam(address, size, pageSize)) {
+                        return;
+                    }
+                    address += pageSize + size;
+                    size = 0;
+                } else {
+                    size += pageSize;
+                }
+            }
+            if (saveInRam(address, size, pageSize)) {
+                return;
+            }
+        }
     }
 
     void writeVersion()
@@ -549,7 +651,9 @@ private:
         shutdown();
     }
 
-    struct LockCheckFailed{};
+    struct LockCheckFailed
+    {
+    };
 
     /**
      * To prevent deadlocks on shutdown, we try to lock from the timer thread
@@ -580,6 +684,11 @@ private:
                 fprintf(stderr, "WARNING: Failed to open /proc/self/statm for reading: %s.\n", strerror(errno));
             }
 
+            procMaps = open("/proc/self/maps", O_RDONLY);
+            if (procMaps == -1) {
+                fprintf(stderr, "WARNING: Failed to open /proc/self/maps for reading: %s.\n", strerror(errno));
+            }
+
             // ensure this utility thread is not handling any signals
             // our host application may assume only one specific thread
             // will handle the threads, if that's not the case things
@@ -597,6 +706,7 @@ private:
             timerThread = thread([&]() {
                 RecursionGuard::isActive = true;
                 debugLog<MinimalOutput>("%s", "timer thread started");
+                int counter = -1;
 
                 // now loop and repeatedly print the timestamp and RSS usage to the data stream
                 while (!stopTimerThread) {
@@ -607,6 +717,10 @@ private:
                         HeapTrack heaptrack([&] { return !stopTimerThread.load(); });
                         heaptrack.writeTimestamp();
                         heaptrack.writeRSS();
+                        counter = (counter + 1) % 100;
+                        if (counter == 0) {
+                            heaptrack.writeInRam();
+                        }
                     } catch (LockCheckFailed) {
                         break;
                     }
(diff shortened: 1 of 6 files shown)


src/track/libheaptrack.cpp
```diff
@@ -549,7 +651,9 @@ class HeapTrack
         shutdown();
     }
 
-    struct LockCheckFailed{};
+    struct LockCheckFailed
```
milianw: unrelated change

src/track/libheaptrack.cpp
```diff
@@ -607,6 +717,10 @@ class HeapTrack
                         HeapTrack heaptrack([&] { return !stopTimerThread.load(); });
                         heaptrack.writeTimestamp();
                         heaptrack.writeRSS();
+                        counter = (counter + 1) % 100;
+                        if (counter == 0) {
+                            heaptrack.writeInRam();
```
milianw: this means that the data is potentially obsolete when the application ends. at the very least you should try to write the ram data when the app is closing down, no?

src/util/pointermap.h
```diff
@@ -61,27 +61,41 @@ struct hash<IndexedAllocationInfo>
 };
 }
 
+//todo change name
```
milianw: remove the comment

src/util/pointermap.h
```diff
@@ -61,27 +61,41 @@ struct hash<IndexedAllocationInfo>
...
+            auto ret = m_set.insert(it, info);
+            m_allocationIndexToInfo.emplace_back(ret);
             return true;
         }
     }
-    std::unordered_set<IndexedAllocationInfo> set;
+
+    bool findAllocationInfo(AllocationInfoIndex index, IndexedAllocationInfo& info)
+    {
+        if(index.index >= m_allocationIndexToInfo.size())
```
milianw: wrap with next line and space after if

src/util/pointermap.h
```diff
@@ -163,6 +177,24 @@ class PointerMap
         return {index, true};
     }
 
+    std::pair<AllocationInfoIndex, uint64_t> takePointer()
+    {
+        AllocationInfoIndex index;
+        if(map.empty())
```
milianw: dito

src/util/pointermap.h
```diff
@@ -163,6 +177,24 @@ class PointerMap
...
+        {
+            return {index, 0};
+        }
+        auto mapIt = map.begin();
+        auto ptr = mapIt->first * SplitPointer::PageSize + mapIt->second.smallPtrParts.back();
+        index = mapIt->second.allocationIndices.back();
+        mapIt->second.allocationIndices.pop_back();
+        mapIt->second.smallPtrParts.pop_back();
+        if(mapIt->second.smallPtrParts.empty())
```
milianw: dito

src/track/libheaptrack.cpp
```diff
@@ -400,6 +405,103 @@ class HeapTrack
...
+
+        debugLog<VeryVerboseOutput>("save in RAM %zx %zx", address / pageSize, size / pageSize);
+        if (!s_data->out.write(" %zx %zx", address / pageSize, size / pageSize)) {
+            return 1;
+        }
+
+        return 0;
+    }
+
+    void writeInRam()
```
milianw: I believe you shouldn't manually parse `/proc/maps`, but rather use `dl_iterate_phdr` instead.

src/track/libheaptrack.cpp
```diff
@@ -276,6 +279,7 @@ class HeapTrack
             pthread_atfork(&prepare_fork, &parent_fork, &child_fork);
 
             atexit([]() {
+                // TODO: it does not work
```
milianw: ? :) then please fix it

src/track/libheaptrack.cpp
tomaszduda23: Shared objects does not matter here(it seems that dl_iterate_phdr list only those). It should list all virtual addresses which are valid. Getting them all seems to be the the fastest way.
The other option would be to list all leaked addresses. There is no easy(fast) way to list all leaked address.

milianw: ah yes right, `dl_iterate_phdr` is useless in this context - sorry for that.

can you maybe give a more motivating example compared to the one given in your commit message? did you actually find interesting issues with this metric that aren't possible to find with the existing messages?

if the additional overhead and complexity is worth it, then we should try to get this in. but as-is, I don't see it being too useful, especially since it's not tracked over time but only correlates to the last sampling time. I.e. if this new metric is useful, then we should also try to sample its data more often to investigate the memory peak which is going to be different from the "heap peak" we collect currently.

I still think that doing that properly would require a different kind of tracer (which could write the same output format though). if we could get our hands on page in/out and swap in/out events with perf, maybe that could be used?

tomaszduda23: I need to reduce memory usage when my app is in the background. Leaked flame graph tab is perfect for that(BTW for memory leaks there is better one. I will send you patch when I have some time) .
I wait when my app goes to background. I dump the data and check who still keeps stuff in memory. The problem is that there are custom memory allocators (e.g. glib - it is possible to disable it but it is not always the case for others) and over allocations. There is a need to check how much memory was touched.

It samples data over time(actually it was not my intention but there was no easy way to do it otherwise). You can argue that it is not precise but it does not matter. Things like flame graphs are not precise. It just show you direction where you should look. The issue was that with current data format I was not able to find a way to pass "data over time" from "interpret" to "analyze" efficiently. Maybe adding additional data layer would be the solution.

Overhead seems to be acceptable. And it is possible to make "In RAM" data optional. Since one needs to understand how lazy memory allocation works in Linux to make sense of it.

Tracing page in/out and swap in/out does not solve the problem. Since you only know data on page level. Let's say you have 4K pages. 1b can make "page in" but the other 4095b may cause memory leak.

Linux perf is great tool but it is completely different league. I was able to setup your tool on production embedded device in a few hours (with a small adjustments). Perf is not that simple. First of all you need kernel which support it (you may get it in production kernel or may not). Than you need backtrace to work. Which may be pain in the ass.
I was looking into that some time ago (it may be different now). On ARM there were 2 ways to get call stack:
- walk using frame pointer - it won't work in many cases. Since there is no reason to care about it. And code can be faster if you don't care.
- copy stack data to perf tool and postmortem use libunwind to get backtrace. It make performance problems. By default it copies 4K of stack which is not enough. When you increase the data amount it make more performance problems. Effort is required to solve those.

milianw: > I need to reduce memory usage when my app is in the background. Leaked flame graph tab is perfect for that(BTW for memory leaks there is better one. I will send you patch when I have some time) .

You mean peak? And what do you mean by "in the background"? When it's idle?

> I wait when my app goes to background. I dump the data and check who still keeps stuff in memory. The problem is that there are custom memory allocators (e.g. glib - it is possible to disable it but it is not always the case for others) and over allocations. There is a need to check how much memory was touched.

I cannot follow you. What do custom memory allocators have to do with this patch? If they are really custom and circumvent `malloc` and use `mmap` directly, heaptrack won't intercept them.

> It samples data over time(actually it was not my intention but there was no easy way to do it otherwise). You can argue that it is not precise but it does not matter. Things like flame graphs are not precise.

How are flame graphs not precise? They are, for what they show. Time sampling is different, in that it can miss peaks which can be hazardous. I.e. if you allocate memory and make it dirty and then free it up, the time sampling may miss it but it's what is actually interesting.

> It just show you direction where you should look. The issue was that with current data format I was not able to find a way to pass "data over time" from "interpret" to "analyze" efficiently. Maybe adding additional data layer would be the solution.

We are already passing data over time for the consumption graphs etc. no?

> Overhead seems to be acceptable. And it is possible to make "In RAM" data optional. Since one needs to understand how lazy memory allocation works in Linux to make sense of it.
> 
> Tracing page in/out and swap in/out does not solve the problem. Since you only know data on page level. Let's say you have 4K pages. 1b can make "page in" but the other 4095b may cause memory leak.

I don't mean tracing only page in/out, but rather doing that in addition to normal heap tracing. The page tracing would make the timer-based sampling of `/proc/maps` superfluous and would allow one to accurately report peak consumptions.

> Linux perf is great tool but it is completely different league. I was able to setup your tool on production embedded device in a few hours (with a small adjustments). Perf is not that simple. First of all you need kernel which support it (you may get it in production kernel or may not). Than you need backtrace to work. Which may be pain in the ass.
> I was looking into that some time ago (it may be different now). On ARM there were 2 ways to get call stack:
>
>     * walk using frame pointer - it won't work in many cases. Since there is no reason to care about it. And code can be faster if you don't care.
> 
>     * copy stack data to perf tool and postmortem use libunwind to get backtrace. It make performance problems. By default it copies 4K of stack which is not enough. When you increase the data amount it make more performance problems. Effort is required to solve those.

You are right in all of what you are saying. You should try out hotspot though, it makes it much simpler and then you can keep the debug symbols only on the host and don't need to push them to the target. And you can stream the perf.data over the network or similar, if I/O is a problem.

tomaszduda23: > > I need to reduce memory usage when my app is in the background. Leaked flame graph tab is perfect for that(BTW for memory leaks there is better one. I will send you patch when I have some time) .
> 
> You mean peak? And what do you mean by "in the background"? When it's idle?
> 
For instance you are making Netflix app for Android. When you play a video you need a lot of memory. When another app is in foreground. Your app go to background. It free most of memory but it keeps some data to be able quickly go to foreground again. If you want to optimize memory usage when your app is in background "leaked" is more useful.
> > I wait when my app goes to background. I dump the data and check who still keeps stuff in memory. The problem is that there are custom memory allocators (e.g. glib - it is possible to disable it but it is not always the case for others) and over allocations. There is a need to check how much memory was touched.
> 
> I cannot follow you. What do custom memory allocators have to do with this patch? If they are really custom and circumvent `malloc` and use `mmap` directly, heaptrack won't intercept them.
> 
Custom memory allocator may use malloc(20*1024*1024). E.g. glib is doing like that. There is no reason to bother with mmap. malloc will do.

> > It samples data over time(actually it was not my intention but there was no easy way to do it otherwise). You can argue that it is not precise but it does not matter. Things like flame graphs are not precise.
> 
> How are flame graphs not precise? They are, for what they show. Time sampling is different, in that it can miss peaks which can be hazardous. I.e. if you allocate memory and make it dirty and then free it up, the time sampling may miss it but it's what is actually interesting.
> 
Flame graphs won't tell you if there is a memory leak or not. You still need to check source code.

I got your point. How would you like to address this? Even if you track page faults you just can see page size (e.g. 4K). There may be multiple malloc per page. How are you going to connet them.

If you are worried that you miss memory peak there is simple solution. Calling free won't release page from memory. You need to tell kernel explicitly to do so. Those function can be hooked and snapshot can be taken just before.

BTW sampling every 100ms still seems to be reasonable. perf tool is doing so when you track CPU usage. It just has to be clear for the user how data are collected.

> > It just show you direction where you should look. The issue was that with current data format I was not able to find a way to pass "data over time" from "interpret" to "analyze" efficiently. Maybe adding additional data layer would be the solution.
> 
> We are already passing data over time for the consumption graphs etc. no?
> 
"Analyze" GUI is not aware of memory addresses. It just knows indexes. "In RAM" size should be written per each malloc (it would be possible to skip some of them). It seems like much more data than currently. It is solvable problem.

> > Overhead seems to be acceptable. And it is possible to make "In RAM" data optional. Since one needs to understand how lazy memory allocation works in Linux to make sense of it.
> > Tracing page in/out and swap in/out does not solve the problem. Since you only know data on page level. Let's say you have 4K pages. 1b can make "page in" but the other 4095b may cause memory leak.
> 
> I don't mean tracing only page in/out, but rather doing that in addition to normal heap tracing. The page tracing would make the timer-based sampling of `/proc/maps` superfluous and would allow one to accurately report peak consumptions.
> 
I would argue if those data will be reliable enough due to granulation (page size).

> > Linux perf is great tool but it is completely different league. I was able to setup your tool on production embedded device in a few hours (with a small adjustments). Perf is not that simple. First of all you need kernel which support it (you may get it in production kernel or may not). Than you need backtrace to work. Which may be pain in the ass.
> > I was looking into that some time ago (it may be different now). On ARM there were 2 ways to get call stack:
> > ```
> > * walk using frame pointer - it won't work in many cases. Since there is no reason to care about it. And code can be faster if you don't care.
> > 
> > * copy stack data to perf tool and postmortem use libunwind to get backtrace. It make performance problems. By default it copies 4K of stack which is not enough. When you increase the data amount it make more performance problems. Effort is required to solve those.
> > ```
> 
> You are right in all of what you are saying. You should try out hotspot though, it makes it much simpler and then you can keep the debug symbols only on the host and don't need to push them to the target. And you can stream the perf.data over the network or similar, if I/O is a problem.

I don't get it. To get backtrace still perf needs to copy entire stack memory, no? It makes performance problem. Is there better way?

milianw:
I'm closing this due to the long period of inactivity

PR #20, opened 2019-08-27
Porting libheaptrack_preload to Android part.1

vnikolai:
Configuring to build on Android, adding option to use of unwind tables instead of libunwind.

### DIFF
--- a/3rdparty/CMakeLists.txt
+++ b/3rdparty/CMakeLists.txt
@@ -2,7 +2,7 @@ if (ECM_FOUND)
     include(ECMEnableSanitizers)
 endif()
 
-if(HEAPTRACK_BUILD_TRACK)
+if (HEAPTRACK_BUILD_BACKTRACE)
     add_subdirectory(libbacktrace)
 endif()
 
--- a/3rdparty/libbacktrace/CMakeLists.txt
+++ b/3rdparty/libbacktrace/CMakeLists.txt
@@ -170,7 +170,6 @@ include_directories (BEFORE
 )
 include_directories (
     auxincl
-    ${LIBUNWIND_INCLUDE_DIR}
 )
 
 add_library (backtrace STATIC
@@ -185,8 +184,6 @@ add_library (backtrace STATIC
     state.c
     sort.c
 )
-
-target_link_libraries(backtrace ${LIBUNWIND_LIBRARY})
 
 set_target_properties (backtrace PROPERTIES
     COMPILE_FLAGS "${CMAKE_SHARED_LIBRARY_C_FLAGS}"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,6 +62,12 @@ option(
   ${HEAPTRACK_BUILD_ANALYZE_DEFAULT}
 )
 
+option(
+  HEAPTRACK_USE_LIBUNWIND
+  "Define preferred unwind functionality - Libunwind as ON and unwind_tables as OFF."
+  ON
+)
+
 set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
 
 if (NOT MSVC)
@@ -102,7 +108,9 @@ file(RELATIVE_PATH LIB_REL_PATH
 set(ECM_ENABLE_SANITIZERS "" CACHE STRING "semicolon-separated list of sanitizers to enable for code that is not injected into client applications")
 
 if (HEAPTRACK_BUILD_TRACK)
-    find_package(Libunwind REQUIRED)
+    if (HEAPTRACK_USE_LIBUNWIND)
+        find_package(Libunwind REQUIRED)
+    endif()
 
     check_cxx_source_compiles(
         "#include <stdio_ext.h>
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,8 +34,10 @@ set_package_properties(Zstd PROPERTIES TYPE RECOMMENDED PURPOSE "Zstandard offer
 
 if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
     set(HEAPTRACK_BUILD_TRACK_DEFAULT ON)
+    set(HEAPTRACK_BUILD_INTERPRET_DEFAULT ON)
 else()
     set(HEAPTRACK_BUILD_TRACK_DEFAULT OFF)
+    set(HEAPTRACK_BUILD_INTERPRET_DEFAULT OFF)
 endif()
 
 option(
@@ -43,6 +45,22 @@ option(
   "Disable this option to skip building the tracker part for heaptrack, e.g. to only build the GUI."
   ${HEAPTRACK_BUILD_TRACK_DEFAULT}
 )
+
+option(
+  HEAPTRACK_BUILD_BACKTRACE
+  "Enable this option to build libbacktrace. It will be enabled implicitly if HEAPTRACK_BUILD_INTERPRET is enabled."
+  OFF
+)
+
+option(
+  HEAPTRACK_BUILD_INTERPRET
+  "Disable this option to skip building the interpret part for heaptrack."
+  ${HEAPTRACK_BUILD_INTERPRET_DEFAULT}
+)
+
+if (HEAPTRACK_BUILD_INTERPRET)
+    set(HEAPTRACK_BUILD_BACKTRACE ON)
+endif()
 
 if (CMAKE_CROSSCOMPILING)
     set(HEAPTRACK_BUILD_ANALYZE_DEFAULT OFF)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -109,6 +109,7 @@ endif()
 # See: https://bugs.kde.org/show_bug.cgi?id=383889
 include(CheckSymbolExists)
 check_symbol_exists(cfree malloc.h HAVE_CFREE)
+check_symbol_exists(valloc stdlib.h HAVE_VALLOC)
 
 set(BIN_INSTALL_DIR "bin")
 set(LIB_SUFFIX "" CACHE STRING "Define suffix of directory name (32/64)")
--- a/src/track/CMakeLists.txt
+++ b/src/track/CMakeLists.txt
@@ -1,6 +1,5 @@
 include_directories(
     ${Boost_INCLUDE_DIRS}
-    ${LIBUNWIND_INCLUDE_DIR}
 )
 
 # heaptrack: bash script to inject/preload
@@ -12,6 +11,17 @@ install(PROGRAMS ${PROJECT_BINARY_DIR}/${BIN_INSTALL_DIR}/heaptrack
     DESTINATION ${BIN_INSTALL_DIR}
 )
 
+if (HEAPTRACK_USE_LIBUNWIND)
+    include_directories(${LIBUNWIND_INCLUDE_DIR})
+
+    add_library(heaptrack_unwind STATIC trace_libunwind.cpp)
+    target_link_libraries(heaptrack_unwind LINK_PRIVATE ${LIBUNWIND_LIBRARY})
+else()
+    add_library(heaptrack_unwind STATIC trace_unwind_tables.cpp)
+endif()
+
+set_property(TARGET heaptrack_unwind PROPERTY POSITION_INDEPENDENT_CODE ON)
+
 # heaptrack_preload: track a newly started process
 add_library(heaptrack_preload MODULE
     heaptrack_preload.cpp
@@ -21,7 +31,7 @@ add_library(heaptrack_preload MODULE
 target_link_libraries(heaptrack_preload LINK_PRIVATE
     ${CMAKE_DL_LIBS}
     ${CMAKE_THREAD_LIBS_INIT}
-    ${LIBUNWIND_LIBRARY}
+    heaptrack_unwind
     rt
 )
 
@@ -44,7 +54,7 @@ add_library(heaptrack_inject MODULE
 target_link_libraries(heaptrack_inject LINK_PRIVATE
     ${CMAKE_DL_LIBS}
     ${CMAKE_THREAD_LIBS_INIT}
-    ${LIBUNWIND_LIBRARY}
+    heaptrack_unwind
     rt
 )
 
--- a/src/track/heaptrack_inject.cpp
+++ b/src/track/heaptrack_inject.cpp
@@ -23,6 +23,7 @@
 #include <cstdlib>
 #include <cstring>
 
+#include <dlfcn.h>
 #include <errno.h>
 #include <link.h>
 #include <malloc.h>
--- a/src/track/libheaptrack.cpp
+++ b/src/track/libheaptrack.cpp
@@ -445,6 +445,16 @@ public:
         s_data->out.writeHexLine('-', reinterpret_cast<uintptr_t>(ptr));
     }
 
+    static bool isPaused()
+    {
+        return s_paused;
+    }
+
+    static void setPaused(bool state)
+    {
+        s_paused = state;
+    }
+
 private:
     static int dl_iterate_phdr_callback(struct dl_phdr_info* info, size_t /*size*/, void* data)
     {
@@ -640,11 +650,16 @@ private:
 
     static std::mutex s_lock;
     static LockedData* s_data;
+
+private:
+    static std::atomic<bool> s_paused;
 };
 
 std::mutex HeapTrack::s_lock;
 HeapTrack::LockedData* HeapTrack::s_data{nullptr};
+std::atomic<bool> HeapTrack::s_paused{false};
 }
+
 extern "C" {
 
 void heaptrack_init(const char* outputFileName, heaptrack_callback_t initBeforeCallback,
@@ -675,9 +690,19 @@ void heaptrack_stop()
     heaptrack.shutdown();
 }
 
+void heaptrack_pause()
+{
+    HeapTrack::setPaused(true);
+}
+
+void heaptrack_resume()
+{
+    HeapTrack::setPaused(false);
+}
+
 void heaptrack_malloc(void* ptr, size_t size)
 {
-    if (ptr && !RecursionGuard::isActive) {
+    if (!HeapTrack::isPaused() && ptr && !RecursionGuard::isActive) {
         RecursionGuard guard;
 
         debugLog<VeryVerboseOutput>("heaptrack_malloc(%p, %zu)", ptr, size);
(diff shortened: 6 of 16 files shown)


CMakeLists.txt
```diff
@@ -97,6 +97,7 @@ endif()
 # See: https://bugs.kde.org/show_bug.cgi?id=383889
 include(CheckSymbolExists)
 check_symbol_exists(cfree malloc.h HAVE_CFREE)
+check_symbol_exists(valloc stdlib.h HAVE_VALLOC)
```
milianw: this should be a separate patch

src/track/libheaptrack.cpp
```diff
@@ -261,15 +261,6 @@ class HeapTrack
         static once_flag once;
         call_once(once, [] {
             debugLog<MinimalOutput>("%s", "doing once-only initialization");
-            // configure libunwind for better speed
```
milianw: if libunwind isn't available on android, I'm OK with making the backtrace backend optional. abstract setup code and introduce separate files, e.g. `backtrace_libunwind.cpp` and `backtrace_unwind_tables.cpp`, then compile in either of them based on a CMake option. 

Here, we'd then call something like

backtrace::setup();

which would do nothing in the unwind tables case, and would call the code you removed for the libunwind scenario

src/track/trace.h
```diff
@@ -21,14 +21,44 @@
...
-#include <libunwind.h>
+#include <unwind.h>
 
 /**
- * @brief A libunwind based backtrace.
+ * @brief A clang unwind-tables based backtrace.
  */
 struct Trace
 {
+private:
```
milianw: move this code into a separate file

src/track/trace.h
```diff
@@ -59,7 +89,7 @@ struct Trace
     bool fill(int skip)
     {
         int size = unw_backtrace(m_data, MAX_SIZE);
```
milianw: backtrace::fill(m_data, MAX_SIZE) or similar, which gets implemented in both, libunwind or your new custom backtrace lib

src/util/config.h.cmake
```diff
@@ -32,5 +32,6 @@
 // cfree() does not exist in glibc 2.26+.
 // See: https://bugs.kde.org/show_bug.cgi?id=383889
 #cmakedefine01 HAVE_CFREE
+#cmakedefine01 HAVE_VALLOC
```
milianw: part of the valloc patch

CMakeLists.txt
```diff
@@ -77,7 +77,7 @@ option(
 set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
 
 if (NOT MSVC)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wpedantic")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wpedantic -funwind-tables")
```
milianw: why? afaik on x86 the unwind tables are there by default, no? is this required for arm? does this mean heaptrack won't work anymore for binaries compiled without unwind tables?

generally -> the backtrace backend must be configurable, I'm reluctant to remove libunwind. having the ability to switch is nice though, and would allow us to profile the different implementations too

tests/auto/CMakeLists.txt
```diff
@@ -13,7 +13,7 @@ include_directories(
 add_definitions(-DCATCH_CONFIG_MAIN)
 
 add_executable(tst_trace tst_trace.cpp)
-#target_link_libraries(tst_trace ${LIBUNWIND_LIBRARY})
+target_compile_options(tst_trace PRIVATE -funwind-tables)
```
milianw: see comments for previous commit, this one here should be squashed

3rdparty/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ if (ECM_FOUND)
     include(ECMEnableSanitizers)
 endif()
 
-if(HEAPTRACK_BUILD_TRACK)
+if(HEAPTRACK_BUILD_INTERPRET)
```
milianw: the code to add the optional BUILD_INTERPRET flag handling should be a separate patch

src/track/CMakeLists.txt
```diff
@@ -1,6 +1,5 @@
 include_directories(
     ${Boost_INCLUDE_DIRS}
-    ${LIBUNWIND_INCLUDE_DIR}
```
milianw: the libunwind handling should be squashed into the previous commit that introduces your separate unwinding backend

tests/auto/CMakeLists.txt
```diff
@@ -6,15 +6,14 @@ set(CMAKE_BUILD_TYPE Debug)
...
-    ${LIBUNWIND_INCLUDE_DIR}
     ${CMAKE_CURRENT_BINARY_DIR}
     ${CMAKE_CURRENT_SOURCE_DIR}/../../
     ${CMAKE_CURRENT_SOURCE_DIR}/../../src
 )
 add_definitions(-DCATCH_CONFIG_MAIN)
 
 add_executable(tst_trace tst_trace.cpp)
-target_link_libraries(tst_trace ${LIBUNWIND_LIBRARY})
+#target_link_libraries(tst_trace ${LIBUNWIND_LIBRARY})
```
milianw: commented out code

vnikolai:
> Hey, thanks a lot for your contribution!
> 
> I'm OK with adding a separate backtrace backend, but you'll have to cleanup the implementation. Please keep libunwind working and on-by-default on non-android platforms. Keep the ifdefs to a minimum, use separate implementation files for the two unwinders and just make a minimal init + backtrace API surface public.
> 
> Then, please also expand your commit messages and talk more about why you are doing something, not just what you are doing.
> 
> Also please rebase your changes to remove the merge commit, and split and squash the commits, such that they form independent units, but everything that belongs together is grouped together.
> 
> Furthermore, please address the following questions in your commits:
> 
> * Why is `-funwind-tables` required explicitly, can't the other unwinder cope with `-fasynchronous-unwind-tables` which is afaik on-by-default? What about support for ARM's exidx which is afaik also mandated by the ABI and thus always available? If this new backtrace backend doesn't support these scenarios and depends on custom compile flags, then we must not use it by default.
> * Please also benchmark the two backtrace implementations, i.e. measure the following for both backends and report the numbers:
> 
> ```
> $ cat run.sh 
> #!/bin/bash
> cd $(dirname $0)
> LD_PRELOAD=$(readlink -f ./lib/heaptrack/libheaptrack_preload.so) DUMP_HEAPTRACK_OUTPUT=/dev/null $@
> $ perf script -r 5 ./run.sh tests/manual/threaded
> ```
> 
> Thanks!

Thanks for fb, appreciate it! I'm wip fixing.

3rdparty/libbacktrace/CMakeLists.txt
```diff
@@ -170,7 +170,6 @@ include_directories (BEFORE
 )
 include_directories (
     auxincl
-    ${LIBUNWIND_INCLUDE_DIR}
```
vnikolai: Was not likely used here.

src/track/trace_libunwind.cpp
```diff
@@ -0,0 +1,30 @@
+/**
```
vnikolai: Should I put any copyright notice here? If so, what it should be exactly?

src/track/trace_unwind_tables.cpp
```diff
@@ -0,0 +1,40 @@
+/**
```
vnikolai: Should I put any copyright notice here? If so, what it should be exactly?

vnikolai:
> > Hey, thanks a lot for your contribution!
> > I'm OK with adding a separate backtrace backend, but you'll have to cleanup the implementation. Please keep libunwind working and on-by-default on non-android platforms. Keep the ifdefs to a minimum, use separate implementation files for the two unwinders and just make a minimal init + backtrace API surface public.
> > Then, please also expand your commit messages and talk more about why you are doing something, not just what you are doing.
> > Also please rebase your changes to remove the merge commit, and split and squash the commits, such that they form independent units, but everything that belongs together is grouped together.
> > Furthermore, please address the following questions in your commits:
> > 
> > * Why is `-funwind-tables` required explicitly, can't the other unwinder cope with `-fasynchronous-unwind-tables` which is afaik on-by-default? What about support for ARM's exidx which is afaik also mandated by the ABI and thus always available? If this new backtrace backend doesn't support these scenarios and depends on custom compile flags, then we must not use it by default.
> > * Please also benchmark the two backtrace implementations, i.e. measure the following for both backends and report the numbers:
> > 
> > ```
> > $ cat run.sh 
> > #!/bin/bash
> > cd $(dirname $0)
> > LD_PRELOAD=$(readlink -f ./lib/heaptrack/libheaptrack_preload.so) DUMP_HEAPTRACK_OUTPUT=/dev/null $@
> > $ perf script -r 5 ./run.sh tests/manual/threaded
> > ```
> > 
> > 
> > Thanks!
> 
> Thanks for fb, appreciate it! I'm wip fixing.

Hi!

Imho I addressed most of the concerns for at the moment, delaying those parts which should be requested separately to reduce review, however no benchmark result yet, wondering if they are critical to the date. Is there any CI instructions to make sure nothing critical missed in my patch?

Thank you very much!

3rdparty/libbacktrace/CMakeLists.txt
milianw: true, so make it a separate commit

CMakeLists.txt
```diff
@@ -62,6 +62,12 @@ option(
   ${HEAPTRACK_BUILD_ANALYZE_DEFAULT}
 )
 
+option(
+  HEAPTRACK_HAVE_LIBUNWIND
```
milianw: the option should be called USE not HAVE

src/track/CMakeLists.txt
```diff
@@ -12,6 +12,15 @@ install(PROGRAMS ${PROJECT_BINARY_DIR}/${BIN_INSTALL_DIR}/heaptrack
...
 )
 
+if (HEAPTRACK_HAVE_LIBUNWIND)
+    add_library(heaptrack_unwind STATIC trace_libunwind.cpp)
+    target_link_libraries(heaptrack_unwind LINK_PRIVATE ${LIBUNWIND_LIBRARY})
+else()
+    add_library(heaptrack_unwind STATIC trace_unwind_tables.cpp)
+endif()
+
+target_compile_options(heaptrack_unwind PRIVATE -fPIC) #some symbols (_Unwind_GetIP etc) require position independent code
```
milianw: this should be moved into the else above

but quite frankly, I don't understand why *using* _Unwind_GetIP requires PIC. Rather, _Unwind_GetIP may not return useful data when unwinding code that wasn't compiled with -fPIC maybe? Can you please expand on this, maybe reference some documentation that says this is required?t

also, the preferred way to achieve this is via
```
set_property(TARGET lib1 PROPERTY POSITION_INDEPENDENT_CODE ON)
```

src/track/trace.h
```diff
@@ -69,6 +64,13 @@ struct Trace
         return m_size > 0;
     }
 
+    static void setup();
+
+    static void print();
+
+private:
+    static int unwind_backtrace(void** data, int max_size);
```
milianw: rename to unwind or fill or just backtrace

src/track/trace_libunwind.cpp
milianw: the same as for the trace.h header but with an updated year (it's all my code basically)

src/track/trace_unwind_tables.cpp
milianw: the same as for trace.h but with your name instead of mine and 2019 as year

src/track/trace_unwind_tables.cpp
```diff
@@ -0,0 +1,44 @@
...
+    int ctr = 0;
+    int max_size = 0;
+};
+
+void Trace::setup()
+{
+}
+
+void Trace::print()
+{
```
milianw: please implement this in the basic form at least:

```
void Trace::print()
{
    Trace trace;
    trace.fill(1);
    for (auto ip : trace)
        fprintf(stderr, "%p\n", ip);
}
```

src/track/trace_unwind_tables.cpp
```diff
@@ -0,0 +1,44 @@
+/**
+ * @brief A unwind-tables based backtrace.
+ */
+
+#include "trace.h"
+
+#include <unwind.h>
+
+struct backtrace
```
milianw: wrap in anon namespace

src/track/trace.h
milianw: also, no need to pass max_size, you can use MAX_SIZE directly

milianw: and rename data to buffer

src/track/trace_unwind_tables.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+void Trace::setup()
+{
+}
+
+void Trace::print()
+{
+}
+
+_Unwind_Reason_Code unwind_backtrace_callback(struct _Unwind_Context* context, void* arg)
```
milianw: move up and also wrap in anon namespace

src/track/CMakeLists.txt
vnikolai: I've learned about this from link error actually:

1. HEAPTRACK_USE_LIBUNWIND - ON

/usr/bin/ld: libheaptrack_unwind.a(trace_libunwind.cpp.o): relocation R_X86_64_32S against symbol `stderr@@GLIBC_2.2.5' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Nonrepresentable section on output
clang: error: linker command failed with exit code 1 (use -v to see invocation)

1. HEAPTRACK_USE_LIBUNWIND - OFF

/usr/bin/ld: libheaptrack_unwind.a(trace_unwind_tables.cpp.o): relocation R_X86_64_PC32 against symbol `_Unwind_GetIP@@GCC_3.0' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Bad value
clang: error: linker command failed with exit code 1 (use -v to see invocation)

I believe it is necessary for building STATIC library with those symbols. Bulding SHARED library implies that by default.
https://cmake.org/cmake/help/v3.0/prop_tgt/POSITION_INDEPENDENT_CODE.html

Making heaptrack_unwind SHARED is an option here.

vnikolai:
> thanks, some more cleanup is required but otherwise this is already acceptable to me
> 
> regarding testing: have you tried running heaptrack with the new unwinder? does it actually still work?

Hi! Thanks once more for clear and accurate fb!
I think I've addressed it as expected atm.

I did tried and somehow succeeded to use customized version of preload library on Android to get the sample report with instruction pointers, however I did use -funwind-tables compile option for my project. I did not recheck it relying on -fasynchronous-unwind-tables and I believe it is not urgent and use-case with client setting -funwind-tables may be a good starting point.

Thanks a lot! ;)

src/track/heaptrack_inject.cpp
```diff
@@ -23,6 +23,7 @@
 #include <cstdlib>
 #include <cstring>
 
+#include <dlfcn.h>
```
vnikolai: for dlopen, dlclose

src/track/libheaptrack.cpp
```diff
@@ -445,6 +445,16 @@ class HeapTrack
         s_data->out.writeHexLine('-', reinterpret_cast<uintptr_t>(ptr));
     }
 
+    static bool isPaused()
+    {
+        return s_paused;
+    }
+
+    static void pause(bool state)
```
milianw: setPaused

src/track/libheaptrack.cpp
```diff
@@ -640,10 +650,14 @@ class HeapTrack
 
     static std::mutex s_lock;
     static LockedData* s_data;
+
+private:
+    static bool s_paused;
```
milianw: `std::atomic<bool>`

src/track/libheaptrack.cpp
vnikolai: > two small changes, and then I'll hope to have time to test and merge this soon - but can't really give an ETA yet on that. hope that's OK with you!
> 
> thanks again for your contributions

Fixed.
Thank you!

vnikolai:
> Hey, thanks a lot for your contribution!
> 
> I'm OK with adding a separate backtrace backend, but you'll have to cleanup the implementation. Please keep libunwind working and on-by-default on non-android platforms. Keep the ifdefs to a minimum, use separate implementation files for the two unwinders and just make a minimal init + backtrace API surface public.
> 
> Then, please also expand your commit messages and talk more about why you are doing something, not just what you are doing.
> 
> Also please rebase your changes to remove the merge commit, and split and squash the commits, such that they form independent units, but everything that belongs together is grouped together.
> 
> Furthermore, please address the following questions in your commits:
> 
>     * Why is `-funwind-tables` required explicitly, can't the other unwinder cope with `-fasynchronous-unwind-tables` which is afaik on-by-default? What about support for ARM's exidx which is afaik also mandated by the ABI and thus always available? If this new backtrace backend doesn't support these scenarios and depends on custom compile flags, then we must not use it by default.
> 
>     * Please also benchmark the two backtrace implementations, i.e. measure the following for both backends and report the numbers:
> 
> 
> ```
> $ cat run.sh 
> #!/bin/bash
> cd $(dirname $0)
> LD_PRELOAD=$(readlink -f ./lib/heaptrack/libheaptrack_preload.so) DUMP_HEAPTRACK_OUTPUT=/dev/null $@
> $ perf script -r 5 ./run.sh tests/manual/threaded
> ```
> 
> Thanks!

Hi!
Not sure if I can make perf script -r work for me, is there any other way to benchmark?
Thanks ;)

vnikolai:
Hello Milian!
Did you have a time to finalize the review?
Thank you very much ;)

milianw:
sorry for the overly long delay, lgtm now - I'll test it locally now and do a quick benchmark and see that nothing else breaks

milianw:
I've tried to push, but KDE needs your full name - can you please tell me it?
otherwise great work, thanks again!

please do note the performance of your implementation though, it's roughly ~5x slower compared to using libunwind

vnikolai:
Hi.
My full name is Volodymyr Nikolaichuk.

Thanks a lot for your time!

Volodymyr

сб, 16 лист. 2019, 21:50 користувач Milian Wolff <notifications@github.com>
пише:

milianw:
pushed now, thanks! please see the commit log of https://github.com/KDE/heaptrack/commit/8023779f9a3be349975bd5b68088b6c0ba0ea0f1 for some perf stat measurements on my machine

thanks again

PR #21, opened 2019-10-09
Declare new hash function in namespace std in order to fix build.

bastianbeischer:
Fixes:

.../src/analyze/gui/parser.cpp:37:8: error: explicit specialization of ‘template<class _Tp> struct std::hash’ outside its namespace must use a nested-name-specifier [-fpermissive]
   37 | struct hash<QString> {
      |        ^~~~~~~~~~~~~

seen with GCC 9.2.0 on Arch Linux.


bastianbeischer:
I've posted this again as:

https://phabricator.kde.org/D24532

I suppose github is just as used as a mirror... Closing.

PR #22, opened 2019-10-14
FIX: Allow high priv user to profile low priv user process

SamSaffron:
In some cases `root` may attach to a non `root` process, in that case
we need to ensure th pipe can be written to by the actual process being
profiled

see: https://bugs.kde.org/show_bug.cgi?id=412742

### DIFF
--- a/src/track/heaptrack.sh.cmake
+++ b/src/track/heaptrack.sh.cmake
@@ -194,6 +194,16 @@ LIBHEAPTRACK_INJECT=$(readlink -f "$LIBHEAPTRACK_INJECT")
 pipe=/tmp/heaptrack_fifo$$
 mkfifo $pipe
 
+# if root is profiling a process for non root
+# give profiled process write access to the pipe
+if [ ! -z "$pid" ]; then
+  pid_user=$(stat -c %U /proc/"$pid")
+  if [ -z "$pid_user" ]; then
+    exit 1
+  fi
+  chown "$pid_user" "$pipe" || exit 1
+fi
+
 output_suffix="gz"
 COMPRESSOR="gzip -c"
 


src/track/heaptrack.sh.cmake
```diff
@@ -194,6 +194,15 @@ LIBHEAPTRACK_INJECT=$(readlink -f "$LIBHEAPTRACK_INJECT")
 pipe=/tmp/heaptrack_fifo$$
 mkfifo $pipe
 
+# if say root is profiling a process for non root
+# give actual process access to the pipe
+if [ ! -z "$pid" ]; then
+  PROCESS_USER=$(ps -o user= "$pid" | xargs)
```
milianw: why do we need the `xargs` call here?

I think I'd prefer this direct approach:

```
if [ ! -z "$pid" ]; then
    pid_user=$(stat -c %U" /proc/"$pid")
    if [ -zu "$pid_user" ]; then
        exit 1;
    fi
    chown "$pid_user" "$pipe" || exit 1;
fi
```
it should be safe to always call this, if the user is the same it doesn't really matter - I don't think we need to optimize

src/track/heaptrack.sh.cmake
```diff
@@ -194,6 +194,15 @@ LIBHEAPTRACK_INJECT=$(readlink -f "$LIBHEAPTRACK_INJECT")
 pipe=/tmp/heaptrack_fifo$$
 mkfifo $pipe
 
+# if say root is profiling a process for non root
+# give actual process access to the pipe
```
milianw: *write* access, everyone already has read access afaik

SamSaffron:
Thanks heaps and no problems, made changes per comments, I thought I needed to trim some spaces which is why I sent stuff to xargs, but it appears it is not needed.

milianw:
merged to master now, thanks a lot!

PR #24, opened 2020-04-01
Add FreeBSD support

myfreeweb:
Wasn't that hard to port, I even got the injection mode running :)

- had to initialize `hooks::calloc` very early to prevent infinite recursion
- had to `flock` a separate lockfile because it's not currently implemented for fifos (I'll try fixing that in the kernel, seems odd, but we have to deal with that on released versions anyway)

### DIFF
--- a/3rdparty/libbacktrace/dwarf.c
+++ b/3rdparty/libbacktrace/dwarf.c
@@ -1854,10 +1854,12 @@ read_line_program (struct backtrace_state *state, struct dwarf_data *ddata,
 		  }
 	      }
 	      break;
+#ifdef DW_LNE_set_discriminator
 	    case DW_LNE_set_discriminator:
 	      /* We don't care about discriminators.  */
 	      read_uleb128 (line_buf);
 	      break;
+#endif
 	    default:
 	      if (!advance (line_buf, len - 1))
 		return 0;
--- a/src/track/heaptrack.sh.cmake
+++ b/src/track/heaptrack.sh.cmake
@@ -228,6 +228,11 @@ cleanup() {
         #       crashes in the debuggee. So instead, we keep heaptrack loaded.
     fi
     rm -f "$pipe"
+    case $(uname) in
+        FreeBSD*)
+            rm -f "$pipe.lock"
+        ;;
+    esac
     kill "$debuggee" 2> /dev/null
 
     echo "Heaptrack finished! Now run the following to investigate the data:"
--- a/src/track/heaptrack.sh.cmake
+++ b/src/track/heaptrack.sh.cmake
@@ -117,7 +117,14 @@ while true; do
                 echo "Missing PID argument."
                 exit 1
             fi
-            client=$(cat /proc/$pid/comm)
+            case $(uname) in
+                Linux*)
+                    client=$(cat "/proc/$pid/comm")
+                ;;
+                FreeBSD*)
+                    client=$(awk '{print $1}' < "/proc/$pid/cmdline")
+                ;;
+            esac
             if [ -z "$client" ]; then
                 echo "Cannot attach to unknown process with PID $pid."
                 exit 1
@@ -197,7 +204,14 @@ mkfifo $pipe
 # if root is profiling a process for non root
 # give profiled process write access to the pipe
 if [ ! -z "$pid" ]; then
-  pid_user=$(stat -c %U /proc/"$pid")
+  case $(uname) in
+    Linux*)
+      pid_user=$(stat -c %U "/proc/$pid")
+    ;;
+    FreeBSD*)
+      pid_user=$(stat -f %Su "/proc/$pid")
+    ;;
+  esac
   if [ -z "$pid_user" ]; then
     exit 1
   fi
@@ -255,17 +269,25 @@ else
         --eval-command="run" --args "$client" "$@"
   else
     echo "injecting heaptrack into application via GDB, this might take some time..."
+    case $(uname) in
+        Linux*)
+            dlopen=__libc_dlopen_mode
+        ;;
+        FreeBSD*)
+            dlopen=dlopen@plt
+        ;;
+    esac
     if [ -z "$debug" ]; then
         gdb --batch-silent -n -iex="set auto-solib-add off" -p $pid \
             --eval-command="sharedlibrary libc.so" \
-            --eval-command="call (void) __libc_dlopen_mode(\"$LIBHEAPTRACK_INJECT\", 0x80000000 | 0x002)" \
+            --eval-command="call (void) '$dlopen'(\"$LIBHEAPTRACK_INJECT\", 0x80000000 | 0x002)" \
             --eval-command="sharedlibrary libheaptrack_inject" \
             --eval-command="call (void) heaptrack_inject(\"$pipe\")" \
             --eval-command="detach"
     else
         gdb --quiet -p $pid \
             --eval-command="sharedlibrary libc.so" \
-            --eval-command="call (void) __libc_dlopen_mode(\"$LIBHEAPTRACK_INJECT\", 0x80000000 | 0x002)" \
+            --eval-command="call (void) '$dlopen'(\"$LIBHEAPTRACK_INJECT\", 0x80000000 | 0x002)" \
             --eval-command="sharedlibrary libheaptrack_inject" \
             --eval-command="call (void) heaptrack_inject(\"$pipe\")"
     fi
--- a/src/track/heaptrack_preload.cpp
+++ b/src/track/heaptrack_preload.cpp
@@ -177,7 +177,15 @@ extern "C" {
 
 /// TODO: memalign, pvalloc, ...?
 
-void* malloc(size_t size) noexcept
+// NOTE: adding noexcept to C functions is a hard error in clang++
+//       (but not even a warning in GCC, even with -Wall)
+#if defined(__GNUC__) && !defined(__clang__)
+#define LIBC_FUN_ATTRS noexcept
+#else
+#define LIBC_FUN_ATTRS
+#endif
+
+void* malloc(size_t size) LIBC_FUN_ATTRS
 {
     if (!hooks::malloc) {
         hooks::init();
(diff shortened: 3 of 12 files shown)


milianw:
woha, awesome! I'll try to find time for a proper review later this week. if you don't hear back, please nag :)

cheers

milianw:
whops, please don't merge this yet, this comment was meant for a single commit: https://github.com/KDE/heaptrack/pull/24/commits/52fbb4f470e01596b569aec537814de93c0ccd6c

3rdparty/libbacktrace/dwarf.c
```diff
@@ -1854,10 +1854,12 @@ read_line_program (struct backtrace_state *state, struct dwarf_data *ddata,
 		  }
 	      }
 	      break;
+#ifdef DW_LNE_set_discriminator
```
milianw, line 1857: upstream this please

src/track/heaptrack.sh.cmake
```diff
@@ -117,7 +117,15 @@ while true; do
                 echo "Missing PID argument."
                 exit 1
             fi
-            client=$(cat /proc/$pid/comm)
+            case $(uname) in
```
milianw, line 120: I think we can always use cmdline, that also exists on linux. but it would include the arguments too, which we probably don't want to output on freebsd either. so something like this is hopefully simpler and portable?
```
client=$(awk -F'\0' '{print $1}' < /proc/$pid/cmdline)
```

src/track/heaptrack.sh.cmake
```diff
@@ -197,7 +205,7 @@ mkfifo $pipe
 # if root is profiling a process for non root
 # give profiled process write access to the pipe
 if [ ! -z "$pid" ]; then
-  pid_user=$(stat -c %U /proc/"$pid")
+  pid_user=$(ls -ld /proc/"$pid" | awk 'NR==1 {print $3}')
```
milianw, line 208: https://www.unix.com/man-page/FreeBSD/1/stat/ - can't we use `stat` on freebsd too, just with a different way to query the `u` user?

src/track/heaptrack.sh.cmake
```diff
@@ -250,17 +258,27 @@ else
         --eval-command="run" --args "$client" "$@"
   else
     echo "injecting heaptrack into application via GDB, this might take some time..."
+    case $(uname) in
+        Linux*)
+            dlopen=__libc_dlopen_mode
+        ;;
+        FreeBSD*)
+            dlopen=dlopen@plt
```
milianw, line 266: just a question: will this work with executables that don't link against `-ldl` directly? on linux at least, the `dlopen` symbol resides in that library so if the executable isn't linked against that, heaptrack would fail. but the mysterious libc symbol is always available, which is why I do this here.

src/track/heaptrack.sh.cmake
```diff
@@ -250,17 +258,27 @@ else
...
+        ;;
+    esac
     if [ -z "$debug" ]; then
         gdb --batch-silent -n -iex="set auto-solib-add off" -p $pid \
             --eval-command="sharedlibrary libc.so" \
-            --eval-command="call (void) __libc_dlopen_mode(\"$LIBHEAPTRACK_INJECT\", 0x80000000 | 0x002)" \
+            --eval-command="set \$dlopen = '$dlopen'" \
```
milianw, line 272: the set isn't needed, you could just pass expand the bash var directly in the line below by removing the `\\` before the `$` no?

src/track/heaptrack.sh.cmake
```diff
@@ -250,17 +258,27 @@ else
...
+            --eval-command="call (void) \$dlopen(\"$LIBHEAPTRACK_INJECT\", 0x80000000 | 0x002)" \
             --eval-command="sharedlibrary libheaptrack_inject" \
             --eval-command="call (void) heaptrack_inject(\"$pipe\")" \
             --eval-command="detach"
     else
         gdb --quiet -p $pid \
             --eval-command="sharedlibrary libc.so" \
-            --eval-command="call (void) __libc_dlopen_mode(\"$LIBHEAPTRACK_INJECT\", 0x80000000 | 0x002)" \
+            --eval-command="set \$dlopen = '$dlopen'" \
```
milianw, line 280: dito

src/track/libheaptrack.cpp
```diff
@@ -189,6 +189,16 @@ int createFile(const char* fileName)
         fprintf(stderr, "ERROR: failed to open heaptrack output file %s: %s (%d)\n", outputFileName.c_str(),
                 strerror(errno), errno);
     } else if (flock(out, LOCK_EX | LOCK_NB) != 0) {
+#ifdef __FreeBSD__
+        // pipes do not support flock, create a regular file
+        auto lockpath = outputFileName + ".lock";
+        auto lockfile = open(lockpath.c_str(), O_CREAT | O_WRONLY | O_CLOEXEC, 0644);
+        debugLog<VerboseOutput>("will lock %s/%p\n", lockpath.c_str(), lockfile);
+        if (flock(lockfile, LOCK_EX | LOCK_NB) == 0) {
+            // leaking the fd seems fine
```
milianw, line 209: but that means the extra lock files are leaked too, no? we need to remove them from the shell script I guess?

myfreeweb:
Upstream libbacktrace has its own definitions of all the constants (as enums) since the initial commit there. Sounds like the libbacktrace copy here is either from before git, or has been modified to use the system dwarf headers.

3rdparty/libbacktrace/dwarf.c
myfreeweb: Upstream libbacktrace has its own definitions of all the constants (as enums) since the initial commit there. Sounds like the libbacktrace copy here is either from before git, or has been modified to use the system dwarf headers.

src/track/heaptrack.sh.cmake
myfreeweb: I did that to make it parse `dlopen@plt` correctly, but seems like I could just put it in quotes like `call 'dlopen@plt'()`

3rdparty/libbacktrace/dwarf.c
milianw: ah - I need to update my copy then. thanks for the heads-up!

src/track/heaptrack.sh.cmake
myfreeweb: On FreeBSD, `dlopen` resides in the `rtld`.

`libc` and `libdl` (the latter being a compatibility dummy pretty much) just weakly reference it.

```
/usr/src % readelf -a /usr/lib/libdl.so.1 | rg dlopen
    11: 0000000000001b70    22 FUNC    WEAK   DEFAULT   12 dlopen@@FBSD_1.0
    12: 0000000000001c40    22 FUNC    WEAK   DEFAULT   12 fdlopen@@FBSD_1.3
/usr/src % readelf -a /lib/libc.so.7 | rg dlopen
0000001cb500  053000000007 R_X86_64_JUMP_SLO 000000000019dd90 dlopen@@FBSD_1.0 + 0
   373: 000000000019dff0    22 FUNC    WEAK   DEFAULT   12 fdlopen@@FBSD_1.3
  1328: 000000000019dd90    22 FUNC    WEAK   DEFAULT   12 dlopen@@FBSD_1.0
  2117: 000000000019e050     8 FUNC    WEAK   DEFAULT   12 _rtld_is_dlopened@@FBSDprivate_1.0
/usr/src % readelf -a /libexec/ld-elf.so.1 | rg dlopen
     4: 000000000000ab30    16 FUNC    GLOBAL DEFAULT   11 fdlopen@@FBSD_1.3
    14: 000000000000a990    14 FUNC    GLOBAL DEFAULT   11 dlopen@@FBSD_1.0
    18: 000000000000ca20   234 FUNC    GLOBAL DEFAULT   11 _rtld_is_dlopened@@FBSDprivate_1.0
```

src/track/heaptrack.sh.cmake
myfreeweb: oh, comm doesn't include arguments?

> `awk -F'\0' '{print $1}'`

Weird: this works with `gawk` (GNU awk) but with FreeBSD awk it only returns the first character. But `awk '{print $1}'` without the custom separator returns the first argument just fine..

milianw: hehe but that doesn't work on linux ;-)

in that case, I'm fine with leaving it as is

src/track/heaptrack.sh.cmake
milianw: cool thanks for the info

src/track/heaptrack.sh.cmake
milianw: yep that sounds better to me

src/track/libheaptrack.cpp
myfreeweb: yep, removing from the script now

myfreeweb:
Updated the script. Also, changed the `noexcept` thing because it's actually about compilers, not libcs, see commit message. (Why did I think that glibc would have an `ifdef __cplusplus` to make functions `noexcept`? :D)

src/track/heaptrack_preload.cpp
```diff
@@ -175,7 +177,15 @@ extern "C" {
 
 /// TODO: memalign, pvalloc, ...?
 
-void* malloc(size_t size) noexcept
+// NOTE: adding noexcept to C functions is a hard error in clang++
```
milianw, line 180: we should probably remove them altogethen then, but I can do that myself in a follow up

adriaandegroot:
This caused build fallout eventually on i386 because Elf32_Sxword isn't defined by the system headers. That's https://github.com/freebsd/freebsd-ports-kde/commit/625db79bbbadfc81b80fe30eefa0f9259f1e2a9c in the ports tree. @milianw I didn't submit that on Invent in the first place because it looks a bit grotty -- FreeBSD #ifdeffery . Anyway, in packaging we carry that patch.

milianw:
@adriaandegroot  please upstream

PR #25, opened 2020-05-30
[README.md]: Fix path for macOS caveats script

martin-juul:
The script does not exist anymore, as it has been renamed to do-caveats.sh in the kde tap.


milianw:
thank you, merged

PR #26, opened 2020-09-29
feat: file progress from file size

catskul:
Basically counts compressed bytes and compares to file size. 

Also attempts to project remaining time by calculating bytes/per second.

Looking for comments.

[~~oldimage~~](https://user-images.githubusercontent.com/375302/94522207-46233c00-01fd-11eb-833b-50ee7160a6c2.png)

Screenshot with message cleaned up:
image

### DIFF
--- a/src/analyze/accumulatedtracedata.cpp
+++ b/src/analyze/accumulatedtracedata.cpp
@@ -64,13 +64,13 @@ class byte_counter {
 public:
     using char_type = char;
     struct category
-        : boost::iostreams::dual_use
+        : boost::iostreams::input
         , boost::iostreams::filter_tag
         , boost::iostreams::multichar_tag
         , boost::iostreams::optimally_buffered_tag
     {};
 
-    uint64_t bytes() const { return bytes_; }
+    uint64_t bytes() const { return m_bytes; }
     std::streamsize optimal_buffer_size() const { return 0; }
 
     template<typename Source>
@@ -79,20 +79,12 @@ public:
         auto const readsize = boost::iostreams::read(src, str, size);
         if (readsize == -1)
             return -1;
-        bytes_ += readsize;
+        m_bytes += readsize;
         return readsize;
     }
 
-    template<typename Sink>
-    std::streamsize write(Sink& sink, const char* str, std::streamsize size)
-    {
-        auto const writesize = boost::iostreams::write(sink, str, size);
-        bytes_ += writesize;
-        return writesize;
-    }
-
 private:
-    uint64_t bytes_ = 0;
+    uint64_t m_bytes = 0;
 };
 
 
--- a/src/analyze/accumulatedtracedata.cpp
+++ b/src/analyze/accumulatedtracedata.cpp
@@ -60,20 +60,27 @@ ostream& operator<<(ostream& out, const Index<Base> index)
 }
 
 // boost's counter filter uses an int for the count which overflows for large streams; so replace it with a work alike.
-class byte_counter {
+class byte_counter
+{
 public:
     using char_type = char;
-    struct category
-        : boost::iostreams::input
-        , boost::iostreams::filter_tag
-        , boost::iostreams::multichar_tag
-        , boost::iostreams::optimally_buffered_tag
-    {};
+    struct category : boost::iostreams::input,
+                      boost::iostreams::filter_tag,
+                      boost::iostreams::multichar_tag,
+                      boost::iostreams::optimally_buffered_tag
+    {
+    };
 
-    uint64_t bytes() const { return m_bytes; }
-    std::streamsize optimal_buffer_size() const { return 0; }
+    uint64_t bytes() const
+    {
+        return m_bytes;
+    }
+    std::streamsize optimal_buffer_size() const
+    {
+        return 0;
+    }
 
-    template<typename Source>
+    template <typename Source>
     std::streamsize read(Source& src, char* str, std::streamsize size)
     {
         auto const readsize = boost::iostreams::read(src, str, size);
@@ -86,8 +93,6 @@ public:
 private:
     uint64_t m_bytes = 0;
 };
-
-
 }
 
 AccumulatedTraceData::AccumulatedTraceData()
@@ -232,7 +237,7 @@ bool AccumulatedTraceData::read(boost::iostreams::filtering_istream& in, const P
     uint64_t lastAllocationPtr = 0;
 
     auto const uncompressedCount = in.component<byte_counter>(0);
-    auto const compressedCount   = in.component<byte_counter>(2);
+    auto const compressedCount = in.component<byte_counter>(2);
 
     parsingState.pass = pass;
     parsingState.reparsing = isReparsing;
--- a/src/analyze/accumulatedtracedata.cpp
+++ b/src/analyze/accumulatedtracedata.cpp
@@ -28,6 +28,8 @@
 #include <boost/iostreams/filter/gzip.hpp>
 #include <boost/iostreams/filtering_stream.hpp>
 
+#include <boost/filesystem.hpp>
+
 #include "util/config.h"
 #include "util/linereader.h"
 #include "util/pointermap.h"
@@ -56,6 +58,44 @@ ostream& operator<<(ostream& out, const Index<Base> index)
     out << index.index;
     return out;
 }
+
+// boost's counter filter uses an int for the count which overflows for large streams; so replace it with a work alike.
+class byte_counter {
+public:
+    using char_type = char;
+    struct category
+        : boost::iostreams::dual_use
+        , boost::iostreams::filter_tag
+        , boost::iostreams::multichar_tag
+        , boost::iostreams::optimally_buffered_tag
+    {};
+
+    uint64_t bytes() const { return bytes_; }
+    std::streamsize optimal_buffer_size() const { return 0; }
+
+    template<typename Source>
+    std::streamsize read(Source& src, char* str, std::streamsize size)
+    {
+        auto const readsize = boost::iostreams::read(src, str, size);
+        if (readsize == -1)
+            return -1;
+        bytes_ += readsize;
+        return readsize;
+    }
+
+    template<typename Sink>
+    std::streamsize write(Sink& sink, const char* str, std::streamsize size)
+    {
+        auto const writesize = boost::iostreams::write(sink, str, size);
+        bytes_ += writesize;
+        return writesize;
+    }
+
+private:
+    uint64_t bytes_ = 0;
+};
+
+
 }
 
 AccumulatedTraceData::AccumulatedTraceData()
@@ -133,12 +173,15 @@ bool AccumulatedTraceData::read(const string& inputFile, const ParsePass pass, b
     const bool isCompressed = isGzCompressed || isZstdCompressed;
     ifstream file(inputFile, isCompressed ? ios_base::in | ios_base::binary : ios_base::in);
 
+    parsingState.fileSize = boost::filesystem::file_size(inputFile);
+
     if (!file.is_open()) {
         cerr << "Failed to open heaptrack log file: " << inputFile << endl;
         return false;
     }
 
     boost::iostreams::filtering_istream in;
+    in.push(byte_counter()); // caution, ::read dependant on filter order
     if (isGzCompressed) {
         in.push(boost::iostreams::gzip_decompressor());
     } else if (isZstdCompressed) {
@@ -149,6 +192,7 @@ bool AccumulatedTraceData::read(const string& inputFile, const ParsePass pass, b
         return false;
 #endif
     }
+    in.push(byte_counter()); // caution, ::read dependant on filter order
     in.push(file);
 
     return read(in, pass, isReparsing);
@@ -195,7 +239,17 @@ bool AccumulatedTraceData::read(istream& in, const ParsePass pass, bool isRepars
     // allocations, i.e. when a deallocation follows with the same data
     uint64_t lastAllocationPtr = 0;
 
+    auto const & filtIn = dynamic_cast<boost::iostreams::filtering_istream &>(in);
+    auto const uncompCount = filtIn.component<byte_counter>(0);
+    auto const compCount   = filtIn.component<byte_counter>(2);
+
     while (timeStamp < filterParameters.maxTime && reader.getLine(in)) {
+        parsingState.pass = pass;
+        parsingState.reparsing = isReparsing;
+        parsingState.compressedByte = compCount->bytes();
+        parsingState.uncompressedByte = uncompCount->bytes();
+        parsingState.timestamp = timeStamp;
+
         if (reader.mode() == 's') {
             if (pass != FirstPass || isReparsing) {
                 continue;
--- a/src/analyze/accumulatedtracedata.cpp
+++ b/src/analyze/accumulatedtracedata.cpp
@@ -239,15 +239,15 @@ bool AccumulatedTraceData::read(istream& in, const ParsePass pass, bool isRepars
     // allocations, i.e. when a deallocation follows with the same data
     uint64_t lastAllocationPtr = 0;
 
-    auto const & filtIn = dynamic_cast<boost::iostreams::filtering_istream &>(in);
-    auto const uncompCount = filtIn.component<byte_counter>(0);
-    auto const compCount   = filtIn.component<byte_counter>(2);
+    auto const & filterIn = dynamic_cast<boost::iostreams::filtering_istream &>(in);
+    auto const uncompressedCount = filterIn.component<byte_counter>(0);
+    auto const compressedCount   = filterIn.component<byte_counter>(2);
 
     while (timeStamp < filterParameters.maxTime && reader.getLine(in)) {
         parsingState.pass = pass;
         parsingState.reparsing = isReparsing;
-        parsingState.compressedByte = compCount->bytes();
-        parsingState.uncompressedByte = uncompCount->bytes();
+        parsingState.compressedByte = compressedCount->bytes();
+        parsingState.uncompressedByte = uncompressedCount->bytes();
         parsingState.timestamp = timeStamp;
 
         if (reader.mode() == 's') {
--- a/src/analyze/accumulatedtracedata.cpp
+++ b/src/analyze/accumulatedtracedata.cpp
@@ -165,7 +165,7 @@ bool AccumulatedTraceData::read(const string& inputFile, const ParsePass pass, b
     const bool isCompressed = isGzCompressed || isZstdCompressed;
     ifstream file(inputFile, isCompressed ? ios_base::in | ios_base::binary : ios_base::in);
 
-    parsingState.fileSize_b = boost::filesystem::file_size(inputFile);
+    parsingState.fileSize = boost::filesystem::file_size(inputFile);
 
     if (!file.is_open()) {
         cerr << "Failed to open heaptrack log file: " << inputFile << endl;
@@ -238,9 +238,9 @@ bool AccumulatedTraceData::read(boost::iostreams::filtering_istream& in, const P
     parsingState.reparsing = isReparsing;
 
     while (timeStamp < filterParameters.maxTime && reader.getLine(in)) {
-        parsingState.readCompressed_b = compressedCount->bytes();
-        parsingState.readUncompressed_b = uncompressedCount->bytes();
-        parsingState.timestamp_ms = timeStamp;
+        parsingState.readCompressedByte = compressedCount->bytes();
+        parsingState.readUncompressedByte = uncompressedCount->bytes();
+        parsingState.timestamp = timeStamp;
 
         if (reader.mode() == 's') {
             if (pass != FirstPass || isReparsing) {
--- a/src/analyze/accumulatedtracedata.cpp
+++ b/src/analyze/accumulatedtracedata.cpp
@@ -242,11 +242,12 @@ bool AccumulatedTraceData::read(boost::iostreams::filtering_istream& in, const P
     auto const uncompressedCount = in.component<byte_counter>(0);
     auto const compressedCount   = in.component<byte_counter>(2);
 
+    parsingState.pass = pass;
+    parsingState.reparsing = isReparsing;
+
     while (timeStamp < filterParameters.maxTime && reader.getLine(in)) {
-        parsingState.pass = pass;
-        parsingState.reparsing = isReparsing;
-        parsingState.compressedByte = compressedCount->bytes();
-        parsingState.uncompressedByte = uncompressedCount->bytes();
+        parsingState.readCompressed_b = compressedCount->bytes();
+        parsingState.readUncompressed_b = uncompressedCount->bytes();
         parsingState.timestamp_ms = timeStamp;
 
         if (reader.mode() == 's') {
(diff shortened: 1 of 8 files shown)


src/analyze/accumulatedtracedata.h
```diff
@@ -172,6 +172,17 @@ struct AccumulatedTraceData
     std::vector<IpIndex> opNewIpIndices;
 
     std::vector<AllocationInfo> allocationInfos;
+
+    struct ParsingState {
+        boost::uintmax_t fileSize = -1;
```
catskul, line 177: oops got to fix this.

src/analyze/accumulatedtracedata.cpp
```diff
@@ -195,7 +239,17 @@ bool AccumulatedTraceData::read(istream& in, const ParsePass pass, bool isRepars
     // allocations, i.e. when a deallocation follows with the same data
     uint64_t lastAllocationPtr = 0;
 
+    auto const & filtIn = dynamic_cast<boost::iostreams::filtering_istream &>(in);
```
milianw, line 242: please don't abbreviate variable names, this could be `filteringIn`

src/analyze/accumulatedtracedata.cpp
```diff
@@ -195,7 +239,17 @@ bool AccumulatedTraceData::read(istream& in, const ParsePass pass, bool isRepars
...
+    auto const uncompCount = filtIn.component<byte_counter>(0);
```
milianw, line 243: uncompressedCounter

src/analyze/accumulatedtracedata.cpp
```diff
@@ -195,7 +239,17 @@ bool AccumulatedTraceData::read(istream& in, const ParsePass pass, bool isRepars
...
+    auto const compCount   = filtIn.component<byte_counter>(2);
```
milianw, line 244: compressedCounter

src/analyze/accumulatedtracedata.h
milianw: yep, and I would prefer if you use uint64_t for the fileSize

src/analyze/gui/parser.cpp
```diff
@@ -644,6 +644,11 @@ void Parser::parse(const QString& path, const QString& diffBase)
 
 void Parser::parseImpl(const QString& path, const QString& diffBase, const FilterParameters& filterParameters)
 {
+    using std::chrono::steady_clock;
```
milianw, line 647: I personally find it easier to use QElapsedTimer for this purpose, could you please leverage that one?

src/analyze/gui/parser.cpp
```diff
@@ -653,7 +658,37 @@ void Parser::parseImpl(const QString& path, const QString& diffBase, const Filte
 
         const auto isReparsing = data == oldData;
 
-        emit progressMessageAvailable(isReparsing ? i18n("reparsing data...") : i18n("parsing data..."));
+        auto parsingMsg = isReparsing ? i18n("reparsing data...") : i18n("parsing data...");
+        emit progressMessageAvailable(parsingMsg);
+
+        const auto numPasses = data->stringCache.diffMode ? 2 : 3;
+        auto updateProgress = [this, numPasses, parsingMsg](std::shared_ptr<const ParserData> const & data, time_point start){
+            auto passCompletion = 1.0*data->parsingState.compressedByte/data->parsingState.fileSize;
```
milianw, line 666: here and below: please put spaces around operators, afaik I even have a .clang-format in the repo that you can use

src/analyze/gui/parser.cpp
```diff
@@ -653,7 +658,37 @@ void Parser::parseImpl(const QString& path, const QString& diffBase, const Filte
...
+            auto totalCompletion = ((data->parsingState.pass + passCompletion)/numPasses);
+            auto spentTime = steady_clock::now() - start;
+            auto spentTime_s = duration_cast<duration<double>>(spentTime).count();
+            auto totalCompletionPerSec = totalCompletion / spentTime_s;
+            auto passCompletionPerSec = passCompletion / (spentTime_s - (data->parsingState.pass/totalCompletionPerSec)) ;
+            auto passRemainingTime_s = (1.0 - passCompletion) / passCompletionPerSec;
+            auto totalRemainingTime_s = (spentTime_s / totalCompletion) * (1.0 - totalCompletion);
+            auto message = QString(
+                parsingMsg
+                    + i18n("\nseconds remaing: ") + QString::number(static_cast<int>(totalRemainingTime_s))
```
milianw, line 676: here and below: please don't do i18n string puzzles, that's hard to translate. instead, do something like this: `i18n("time remaining: %1", Util::formatTime(totalRemainingTime))`

also note the typo in "remaining" in the text

src/analyze/gui/parser.cpp
```diff
@@ -653,7 +658,37 @@ void Parser::parseImpl(const QString& path, const QString& diffBase, const Filte
...
-        emit progressMessageAvailable(isReparsing ? i18n("reparsing data...") : i18n("parsing data..."));
+        auto parsingMsg = isReparsing ? i18n("reparsing data...") : i18n("parsing data...");
+        emit progressMessageAvailable(parsingMsg);
+
+        const auto numPasses = data->stringCache.diffMode ? 2 : 3;
+        auto updateProgress = [this, numPasses, parsingMsg](std::shared_ptr<const ParserData> const & data, time_point start){
+            auto passCompletion = 1.0*data->parsingState.compressedByte/data->parsingState.fileSize;
+            auto totalCompletion = ((data->parsingState.pass + passCompletion)/numPasses);
+            auto spentTime = steady_clock::now() - start;
+            auto spentTime_s = duration_cast<duration<double>>(spentTime).count();
```
milianw, line 669: please leverage Util::formatTime

src/analyze/gui/parser.cpp
```diff
@@ -653,7 +658,37 @@ void Parser::parseImpl(const QString& path, const QString& diffBase, const Filte
...
+            auto totalRemainingTime_s = (spentTime_s / totalCompletion) * (1.0 - totalCompletion);
+            auto message = QString(
+                parsingMsg
+                    + i18n("\nseconds remaing: ") + QString::number(static_cast<int>(totalRemainingTime_s))
+                    + i18n("\n%/s: ") + QString::number(totalCompletionPerSec*100)
+                    + i18n("\npass #:   ") +  QString::number(data->parsingState.pass + 1)
+                    + i18n("/ ") + QString::number(numPasses)
+                    + i18n("\ncurrent pass %: ") + QString::number(static_cast<int>(passCompletion * 100))
+                    + i18n("\npass seconds remaing: ") + QString::number(static_cast<int>(passRemainingTime_s))
+                    + i18n("\npass MiB(comp): ") + QString::number(data->parsingState.compressedByte/(1024*1024))
```
milianw, line 682: here and below: leverage Util::formatBytes please

src/analyze/gui/parser.cpp
```diff
@@ -670,9 +705,20 @@ void Parser::parseImpl(const QString& path, const QString& diffBase, const Filte
             }
             data->diff(diffData);
             data->stringCache.diffMode = true;
-        } else if (!data->read(stdPath, isReparsing)) {
-            emit failedToOpen(path);
-            return;
+        } else {
+            auto start = steady_clock::now();
```
milianw, line 709: please don't open a thread just for progress reporting, instead add a callback to the parser interface and call it regularly (e.g. every 1000 lines or similar).

src/analyze/gui/parser.cpp
```diff
@@ -706,17 +754,27 @@ void Parser::parseImpl(const QString& path, const QString& diffBase, const Filte
...
         if (!data->stringCache.diffMode) {
             // only build charts when we are not diffing
-            *parallel << make_job([this, data, stdPath, isReparsing]() {
+            *parallel << make_job([this, data, stdPath, isReparsing, updateProgress]() {
                 // this mutates data, and thus anything running in parallel must
                 // not access data
+                emit progress(0);
                 data->prepareBuildCharts();
-                data->read(stdPath, AccumulatedTraceData::ThirdPass, isReparsing);
+                auto read = async(launch::async, [&data, stdPath, isReparsing]() {
```
milianw, line 762: dito

milianw (changes requested):
cool, in general I like where this is going, but there's quite a bit of cleanup required first.

also, can we talk about what data to actually show? your screenshot is pretty noisy, I wonder if we should really show all of that? For the end user I think this shows too much internal detail. For a developer working on heaptrack itself it could be useful, but even there I wonder if I need all of these values.

My suggestion would be something like this:

a) below the progress bar, just show e.g. `i18n("parsing data... ETA: %1", Util::formatTime(remaining))`
b) if you want to keep the other values, split them out into a html table and show that in a separate QLabel in the bottom right corner of the page. This way the labels can be right-aligned and the values left-aligned and don't jump around while the values are changing

src/analyze/accumulatedtracedata.cpp
catskul: Will fix.

By the way, looking at this I'm realizing the interface in theory allows someone to pass in a non-`filtering_istream` in theory, and I'm completely ignoring that with this `dynamic_cast` Is what I'm doing here acceptable to you?

I could imagine either:

 * switching the `dynamic_cast` to a pointer cast and throwing the byte counts inside an `if`-block to still allow for non-`filtering_istream` calls
* switching the function signature to only accept `filtering_istream`. 

thoughts?

src/analyze/accumulatedtracedata.h
catskul: happy to change it. That type was selected because of boost::filesystem uses that type, though that will not change any time soon, and I can't imagine overflow could ever happen here.

On that subject, if we allow/require c++17 or "::experimental" namespace we could avoid boost here.

Also, would you rather the other counts be `uint64_t` and just initalized to 0? I considered `boost::optional`s but wasn't sure if that was overkill.

I'm always torn on whether to use unsigned types. On the one hand these numbers should never really be negative, but on the other hand unsigned types can wreck havoc when they're invariably eventually used in signed math and int64 is so large the extra bit "wasted" on the sign is irrelevant especially compared to the ease of use when needing to mix signed and unsigned arithmetic. 

IIRC I think the c++ standards committee has started to move toward using signed types in more places to avoid the signed/unsigned math conflicts.

src/analyze/gui/parser.cpp
catskul: Yep. I'm still learning some parts of the Qt ecosystem. I'll read up and switch it.

src/analyze/gui/parser.cpp
catskul: This was a result of not quite understanding how `i18n` was used, and slowly expanding what was there. 

Will change.

src/analyze/accumulatedtracedata.cpp
catskul: I went with changing the interface, but happy to change to the run-time solution if you prefer that.

src/analyze/accumulatedtracedata.h
milianw: I don't want to require c++17 yet - this makes it sometimes a bit harder to compile heaptrack for embedded systems (yes, even in 2020...). so please stick to boost for now.

regarding the signedness - :shrug: uint + 0 sounds good to me, but if you ever encounter API that returns -1 to signal an error then better stick to int + -1 or similar

src/analyze/accumulatedtracedata.cpp
```diff
@@ -195,7 +239,16 @@ bool AccumulatedTraceData::read(istream& in, const ParsePass pass, bool isRepars
     // allocations, i.e. when a deallocation follows with the same data
     uint64_t lastAllocationPtr = 0;
 
+    auto const uncompressedCount = in.component<byte_counter>(0);
+    auto const compressedCount   = in.component<byte_counter>(2);
+
     while (timeStamp < filterParameters.maxTime && reader.getLine(in)) {
+        parsingState.pass = pass;
```
milianw, line 246: pass and reparsing only need to be set once above the loop

timestamp only needs to be updated when it changes below in the 'c' conditional. actually, I would even go as far as updating the bytes-read counter there too

src/analyze/accumulatedtracedata.h
```diff
@@ -172,6 +173,17 @@ struct AccumulatedTraceData
     std::vector<IpIndex> opNewIpIndices;
 
     std::vector<AllocationInfo> allocationInfos;
+
+    struct ParsingState {
+        int64_t fileSize_b = 0;
+        int64_t compressedByte = 0;
+        int64_t uncompressedByte = 0;
+        int64_t timestamp_ms = 0;
```
milianw, line 181: just timeStamp

src/analyze/gui/parser.cpp
milianw: I think you should move the progress reporting into the existing handleTimeStamp function and then emit the signal from there

milianw (changes requested):

src/analyze/accumulatedtracedata.cpp
```diff
@@ -56,6 +58,44 @@ ostream& operator<<(ostream& out, const Index<Base> index)
...
+    {
+        auto const readsize = boost::iostreams::read(src, str, size);
+        if (readsize == -1)
+            return -1;
+        bytes_ += readsize;
+        return readsize;
+    }
+
+    template<typename Sink>
+    std::streamsize write(Sink& sink, const char* str, std::streamsize size)
```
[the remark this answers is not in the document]
catskul: Not sure. I didn't completely understand the requirements of such a filter for iostream_filter, but I tried to match the original fairly closely so I didn't break something in misunderstanding a subtlety.

I can try removing it and see what happens.

src/analyze/gui/parser.cpp
catskul: will do. I'll just have to figure out where to put the timer.

I might could put it in the `ParserState` though the logic could get a bit hairy given that a variable number of passes can happen.

catskul: I also noticed:
 * `handleTimeStamp` only gets called on the first pass. I could change that logic, but am worried that has negative consequences.
 * `handleTimeStamp` doesn't currently have access to the parser which has the signals for `progress*` stuff. Not sure if passing that access in is the right thing to do, or to create new signals.

src/analyze/accumulatedtracedata.h
catskul: Will change if you feel strongly about it, but I've been bitten previously so many times in my career by doing math with implicit & mismatched units that I'd resolved to always include units where possible either via type safety or if not then by var name.

src/analyze/gui/parser.cpp
milianw: Suggestion:

- take a progress callback (`std::function` e.g.) as argument in `ParserData()` ctor and store it in a member var
- also store the QElapsedTimer there, and start it from `Parser::parseImpl`
- always call handleTimeStamp and pass the `pass` as an argument (no pun intended)
- then call the callback and return early after it was called for anything but the first pass
- in `heaptrack_print.cpp` just return early if the pass isn't the first pass
- then you can pass a lambda to the `ParserData` from within `Parser::parseImpl` to do the progress tracking as you see fit. This should also make it hopefully easy to do proper progress reporting for the diff case, where we want to aggregate the two runs somehow

src/analyze/accumulatedtracedata.h
milianw: Type safety would be fine - but we only have that for time via `std::chrono` so far...

Anyhow, the rest of the code base operates on ms timestamps anyways (cf. `Util::formatTime`) and bytes (`Util::formatBytes`) and I don't have that unit stored anywhere else in the variable name. So I'd prefer if this stays uniform.

If you'd like, I'd be OK to change the timestamps everywhere to chrono in a follow-up patch, but that shouldn't be done within this one here as it's not directly related.

src/analyze/gui/parser.cpp
catskul: I'll try to get to this tonight (US East Coast).

src/analyze/accumulatedtracedata.cpp
```diff
@@ -56,6 +58,44 @@ ostream& operator<<(ostream& out, const Index<Base> index)
...
+    {
+        auto const writesize = boost::iostreams::write(sink, str, size);
+        bytes_ += writesize;
+        return writesize;
+    }
+
+private:
+    uint64_t bytes_ = 0;
```
milianw, line 95: `m_bytes` is the format used in heaptrack

milianw:
ping @catskul - any update on this one?

cheers

src/analyze/accumulatedtracedata.cpp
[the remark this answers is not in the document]
catskul: Fixed.

src/analyze/accumulatedtracedata.cpp
catskul: done.

src/analyze/accumulatedtracedata.h
catskul: done.

src/analyze/accumulatedtracedata.h
```diff
@@ -172,6 +173,17 @@ struct AccumulatedTraceData
     std::vector<IpIndex> opNewIpIndices;
 
     std::vector<AllocationInfo> allocationInfos;
+
+    struct ParsingState {
+        int64_t fileSize_b = 0;
```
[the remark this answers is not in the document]
catskul: done.

src/analyze/gui/parser.cpp
catskul: Take a look at my latest commits that refactor according to the above spec and see how you feel. 

After looking at the complexity of my refactor, It occurs to me that the code with the new thread might be simpler, and was symmetric to the code above it, which itself opens an async job/thread for reading the diffData.

If my refactor attempt ended up too complex, I'm happy to change back to `std::async` based on your assessment.

catskul: Hmm I seem to have broken the charts in the effort to refactor to the suggestion. Investigating...

catskul:
The git history is a bit of a mess, but I wanted to preserve it until we're settled. 

Any particular preference on how I clean it up? I would imagine the current work would probably represent 3-4 commits, but I can squash it to a single commit if you prefer.

src/analyze/gui/parser.cpp
```diff
@@ -647,16 +658,37 @@ void Parser::parseImpl(const QString& path, const QString& diffBase, const Filte
...
+        auto parsingMsg = isReparsing ? i18n("reparsing data") : i18n("parsing data");
+        emit progressMessageAvailable(parsingMsg);
+
+        auto updateProgress = [this, parsingMsg](const ParserData & data){
+            const auto numPasses = data.stringCache.diffMode ? 2 : 3;
+            auto passCompletion = 1.0 * data.parsingState.readCompressedByte / data.parsingState.fileSize;
+            auto totalCompletion = (data.parsingState.pass + passCompletion)/numPasses;
+            auto spentTime_s = data.parseTimer.elapsed() / 1000.0; // elapsed is in ms
+            auto totalRemainingTime_s = (spentTime_s / totalCompletion) * (1.0 - totalCompletion);
+            auto message = QString(
```
milianw, line 671: this is still a string puzzle that's hard to localize. better use something like this here:

```
auto message = i18n("%1, pass: %2/%3, elapsed: %4, remaining: %5",
    parsingMsg, data.parsingState.pass + 1, numPasses,
    Util::formatTime(spentTime),
    Util::formatTime(totalRemainingTime));
```

src/analyze/gui/parser.cpp
```diff
@@ -647,16 +658,37 @@ void Parser::parseImpl(const QString& path, const QString& diffBase, const Filte
...
     stream() << make_job([this, oldData, path, diffBase, filterParameters]() {
+        const auto isReparsing = (path == m_path && oldData && diffBase.isEmpty());
+        auto parsingMsg = isReparsing ? i18n("reparsing data") : i18n("parsing data");
+        emit progressMessageAvailable(parsingMsg);
+
+        auto updateProgress = [this, parsingMsg](const ParserData & data){
+            const auto numPasses = data.stringCache.diffMode ? 2 : 3;
+            auto passCompletion = 1.0 * data.parsingState.readCompressedByte / data.parsingState.fileSize;
+            auto totalCompletion = (data.parsingState.pass + passCompletion)/numPasses;
+            auto spentTime_s = data.parseTimer.elapsed() / 1000.0; // elapsed is in ms
```
milianw, line 669: the comment is wrong, and why convert from ms to s and then back again below? this should just be

```
auto spentTime = data.parseTimer.elapsed()
```

src/analyze/gui/parser.cpp
```diff
@@ -647,16 +658,37 @@ void Parser::parseImpl(const QString& path, const QString& diffBase, const Filte
...
     using namespace ThreadWeaver;
     stream() << make_job([this, oldData, path, diffBase, filterParameters]() {
+        const auto isReparsing = (path == m_path && oldData && diffBase.isEmpty());
+        auto parsingMsg = isReparsing ? i18n("reparsing data") : i18n("parsing data");
+        emit progressMessageAvailable(parsingMsg);
+
+        auto updateProgress = [this, parsingMsg](const ParserData & data){
+            const auto numPasses = data.stringCache.diffMode ? 2 : 3;
+            auto passCompletion = 1.0 * data.parsingState.readCompressedByte / data.parsingState.fileSize;
+            auto totalCompletion = (data.parsingState.pass + passCompletion)/numPasses;
```
milianw, line 668: spaces around the `/` operator please

src/analyze/gui/parser.cpp
```diff
@@ -647,16 +658,37 @@ void Parser::parseImpl(const QString& path, const QString& diffBase, const Filte
     auto oldData = std::move(m_data);
     using namespace ThreadWeaver;
     stream() << make_job([this, oldData, path, diffBase, filterParameters]() {
+        const auto isReparsing = (path == m_path && oldData && diffBase.isEmpty());
+        auto parsingMsg = isReparsing ? i18n("reparsing data") : i18n("parsing data");
+        emit progressMessageAvailable(parsingMsg);
+
+        auto updateProgress = [this, parsingMsg](const ParserData & data){
```
milianw, line 665: remove space before `&` please

milianw: afaik there's a `.clang-format` you could try to use?

milianw (commented):
really great work, I like the current state - just some really minor nit picks, then this is ready to go in

many thanks!

src/analyze/gui/parser.cpp
catskul: that was leftover from before I was using `formatTime` I think. Fixed.

src/analyze/gui/parser.cpp
catskul: Done. (ran clang-format-diff to avoid formatting the whole file as there were unrelated items that ended up formatted.)

catskul:
Should I clean up the git history?

milianw:
> Should I clean up the git history?

Ideally just squash it all, yes

milianw:
> > Should I clean up the git history?
> 
> Ideally just squash it all, yes

but we can also just do that directly when merging, but I'd like to have a meaningful commit message for the squashed commit then - so either you provide that to me, or you just squash it all yourself

catskul:
> but we can also just do that directly when merging, but I'd like to have a meaningful commit message for the squashed commit then 

Works for me:

"Add progress reporting by counting bytes through a stream filter and comparing with file size" 

should do it, but feel free to change as you see fit.

milianw:
thank you, I've now merged this and also added a simple rate-control as otherwise the updates would come in far too often, taking away CPU from the actual useful task

milianw:
haha and my rate control is crashy :facepalm: - will fix it :)

catskul:
Yeah, I had rate control in the async version, but didn't think to add it in the sync version.

PR #27, opened 2020-11-05
Add symbol/binary path support for heaptrack_gui kchart part

keqiaow:
This is to provide both symbol and binary path support for heaptrack_gui kchart part.

TODO:
Add legend for each chart model


milianw:
I've merged this now (in a slightly cleaned up form) - many thanks!

for the legend, have a look at https://github.com/KDE/massif-visualizer/blob/master/app/charttab.cpp e.g.

please make sure that the legend can be toggled on/off easily (i.e. via context menu, view menu and some shortcut)

PR #28, opened 2020-12-01
Update homebrew dependencies

arr2036:
Looks like the KDE homebrew tap maintainers switched hosting about 5 months back and started depending on more upstream formulae.  This fixes the tap command and updates the names of the dependencies.


milianw:
thank you, merged now

PR #29, opened 2020-12-25
Backported fix in handling ".note.gnu.build-id" section

bezkrovatki:
Build-ID based standalone debug info might be ignored for some shared objects. See details in https://gcc.gnu.org/bugzilla/show_bug.cgi?id=86198


milianw:
cool, merged. that probably means we should update the full copy of libbacktrace in time

cheers

PR #30, opened 2021-04-03
feat: output file name under -o can include %h and %g

itsjacob:


tsdgeos:
Thanks for your contribution :smiley:

This repository is a mirror of a KDE repository. This means that developers are not looking at pull requests created in GitHub, so I'm closing this pull request (actually a bot is doing it).
Please see https://community.kde.org/Infrastructure/Github_Mirror for details on how to contribute to this and other KDE projects.

src/track/heaptrack.sh.cmake
```diff
@@ -100,11 +106,21 @@ while true; do
                 echo "Missing output argument."
                 exit 1
             fi
-            output=$(readlink -f $2)
+            output=$2
+            if [ -n "$(echo $output | grep '%h')" ]; then
```
milianw, line 110: the `grep` isn't needed, is it? you could just replace via `sed` directly, dito below

src/track/heaptrack.sh.cmake
```diff
@@ -100,11 +106,21 @@ while true; do
...
+              output=$(echo $output | sed "s/%h/$(hostname)/g")
+            fi
+            if [ -n "$(echo $output | grep '%g')" ]; then
```
milianw, line 113: `%g` isn't documented, should this be `%p` instead?

milianw (commented):
you can also fixup the commits and push again, I'll then manually apply the changes over on https://invent.kde.org/sdk/heaptrack/-/merge_requests

src/track/heaptrack.sh.cmake
itsjacob: Yes, I should have written %p.

src/track/heaptrack.sh.cmake
itsjacob: Yes, you are correct, the conditional can be removed.

PR #31, opened 2021-04-06
Fix pull request #30

itsjacob:
- use %p instead of %g
- remove redundant conditionals


milianw:
thank you, I merged this code now into master

PR #32, opened 2021-05-21
Update libbacktrace to support DWARF info from newer gcc

zagto:
Heaptrack with the currently included libbacktrace does not work with the DWARF information generated by default on Arch Linux (GCC 11.1.0). Running a program under Heaptrack will produce messages like this:

```
[removed]/heaptrack_interpret.cpp:351 ERROR:Failed to create backtrace state for module [removed]/libheaptrack_preload.so: unrecognized DWARF version in .debug_info at 6 / Success (error code 0)
[removed]/heaptrack_interpret.cpp:351 ERROR:Failed to create backtrace state for module /usr/lib/libstdc++.so.6: unrecognized DWARF version in .debug_info at 6 / Success (error code 0)
[removed]/heaptrack_interpret.cpp:351 ERROR:Failed to create backtrace state for module /usr/lib/libgcc_s.so.1: unrecognized DWARF version in .debug_info at 6 / Success (error code 0)
```

and the gathered information is empty:
```
$ heaptrack_print /home/zagto/heaptrack.ls.50832.zst
reading file "/home/zagto/heaptrack.ls.50832.zst" - please wait, this might take some time...
finished reading file, now analyzing data:

MOST CALLS TO ALLOCATION FUNCTIONS

PEAK MEMORY CONSUMERS

MOST TEMPORARY ALLOCATIONS

total runtime: 0.001000s.
calls to allocation functions: 0 (0/s)
temporary memory allocations: 0 (0/s)
peak heap memory consumption: 0B
peak RSS (including heaptrack overhead): 0B
total memory leaked: 0B
```

Updating the libbacktrace library fixes this problem and Heaptrack seems to work as expected.

I used the diff between the following revisions of gcc to update the libbacktrace files:
- 782e57f2c0900f3c3bbaec4b367568b6d05236b8 (now)
- 42f7424573fbf889f9857ace32bf8bb364bb2146 (old)

### DIFF
--- a/3rdparty/libbacktrace/ChangeLog
+++ b/3rdparty/libbacktrace/ChangeLog
@@ -1,3 +1,935 @@
+2021-05-03  H.J. Lu  <hjl.tools@gmail.com>
+
+	PR bootstrap/99703
+	* configure: Regenerated.
+
+2021-03-03  Ian Lance Taylor  <iant@golang.org>
+
+	* dwarf.c (read_line_program): Don't special case file 0.
+	(read_function_entry): Likewise.
+
+2021-03-02  Ian Lance Taylor  <iant@golang.org>
+
+	PR libbacktrace/98818
+	* dwarf.c (dwarf_buf_error): Add errnum parameter.  Change all
+	callers.
+	* backtrace.h: Update backtrace_error_callback comment.
+
+2021-02-12  Ian Lance Taylor  <iant@golang.org>
+
+	* configure.ac: Check for objcopy --add-gnu-debuglink by using
+	objcopy --help.
+	* configure: Regenerate
+
+2021-01-18  Ian Lance Taylor  <iant@golang.org>
+
+	* Makefile.am (%_dwz): If dwz fails, use uncompressed debug info.
+	* Makefile.in: Regenerate.
+	* configure: Regenerate.
+
+2021-01-18  Ian Lance Taylor  <iant@golang.org>
+
+	PR debug/98716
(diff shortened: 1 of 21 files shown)


milianw:
thank you, this looks great! I just started looking into this last week but didn't finish to rebase our custom heaptrack specific changes. It seems like you ported these really well too - many thanks. I'll test this sometime later this week and merge it then. If you don't hear back from me in time, please shout out here.

Thanks again

zagto:
Small update: in Debug builds I get an additional unwanted item at the top of stack traces, 4 in total instead of 3:
- `Trace::unwind(void**)`
- `Trace::fill(int)`
- `heaptrack_malloc`
- `malloc`
the lastest commit fixes this by ignoring an addtional one

milianw:
> Small update: in Debug builds I get an additional unwanted item at the top of stack traces, 4 in total instead of 3:
> 
>     * `Trace::unwind(void**)`
> 
>     * `Trace::fill(int)`
> 
>     * `heaptrack_malloc`
> 
>     * `malloc`
>       the lastest commit fixes this by ignoring an addtional one

malloc is _not_ unwanted, it should remain in the trace - please remove this one again

zagto:
alright i'll remove it. But then release builds are broken because they don't show it?

milianw:
Oh you are right! I just checked and indeed we don't show 'malloc' and thus it should be removed. I'll cherry-pick your patch and submit it. This patch here also seems to work fine, so I'll merge it too (but I'll squash the two commits, as otherwise bisecting won't work anymore)

Thanks!

PR #33, opened 2021-05-29
track: Check for unw_set_caching_policy before using

kraj:
llvm libunwind does not implement unw_cache_* functions yet
Include inttypes.h got PRI* macros

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/src/track/trace_libunwind.cpp
+++ b/src/track/trace_libunwind.cpp
@@ -26,6 +26,7 @@
 
 #define UNW_LOCAL_ONLY
 #include <libunwind.h>
+#include <inttypes.h>
 
 #include <stdio.h>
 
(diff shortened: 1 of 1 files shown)


src/track/trace_libunwind.cpp
```diff
@@ -60,9 +61,11 @@ void Trace::print()
 void Trace::setup()
 {
     // configure libunwind for better speed
+#if UNW_CACHE_PER_THREAD
     if (unw_set_caching_policy(unw_local_addr_space, UNW_CACHE_PER_THREAD)) {
         fprintf(stderr, "WARNING: Failed to enable per-thread libunwind caching.\n");
     }
+#endif
 #if LIBUNWIND_HAS_UNW_SET_CACHE_SIZE
```
milianw, line 69: this here is set via `cmake/FindLibunwind.cmake` - you'll have to add the same check for `UNW_CACHE_PER_THREAD` and then add it to `libunwind_config.h.cmake`

milianw (changes requested):

milianw:
Interesting, I've never used heaptrack with any other libunwind than the one over at https://github.com/libunwind/libunwind/ - can you tell me how it performs with llvm's libunwind? how do you use it from cmake? if it works well, I'd like to officially document that.

kraj:
> Interesting, I've never used heaptrack with any other libunwind than the one over at https://github.com/libunwind/libunwind/ - can you tell me how it performs with llvm's libunwind? how do you use it from cmake? if it works well, I'd like to officially document that.

so far in my little testing I have not seen any issues. but then I have only run one sample.

PR #34, opened 2021-08-15
Bug 379660 Remove template parameters from flamegraph view

khReichel:
Added a textfilter which removes the long template parameters as
described in the Bug ticket Bug 379660.
Added unit-tests for filtering the entries and removing the template parameters

### DIFF
--- a/src/analyze/gui/flamegraph.cpp
+++ b/src/analyze/gui/flamegraph.cpp
@@ -45,6 +45,7 @@
 
 #include "resultdata.h"
 #include "util.h"
+#include "middleelide.h"
 
 enum CostType
 {
@@ -192,10 +193,12 @@ void FrameGraphicsItem::paint(QPainter* painter, const QStyleOptionGraphicsItem*
         Q_UNREACHABLE();
     }();
 
+    QString middleElidedLabel = MiddleElide::elideAngleBracket(label);
+
     const int height = rect().height();
     painter->drawText(margin + rect().x(), rect().y(), width, height,
                       Qt::AlignVCenter | Qt::AlignLeft | Qt::TextSingleLine,
-                      option->fontMetrics.elidedText(label, Qt::ElideRight, width));
+                      option->fontMetrics.elidedText(middleElidedLabel, Qt::ElideRight, width));
 
     if (m_searchMatch == NoMatch) {
         painter->setPen(oldPen);
--- a/src/analyze/gui/middleelide.cpp
+++ b/src/analyze/gui/middleelide.cpp
@@ -1,0 +1,42 @@
+#include "middleelide.h"
+#include <QChar>
+
+MiddleElide::MiddleElide()
+{
+
+}
+
+QString MiddleElide::elideAngleBracket(const QString& s)
+{
+   return substituteAngleBrackets(s);
+}
+
+QString MiddleElide::substituteAngleBrackets(const QString& s)
+{
+    static QChar startBracket = QChar(u'<');
+    static QChar stopBracket  = QChar(u'>');
+
+    int level = 0;
+    QString result;
+    for (int i=0, n = s.length(); i < n; i++)  {
+        QChar currentChar = s[i];
+        if (currentChar == QChar(startBracket) && level == 0) {
+            result += QChar(startBracket);
+            level++;
+        }
+        else if (currentChar == QChar(startBracket) && level > 0) {
+            level++;
+        }
+        else if (currentChar == QChar(stopBracket) && level == 1) {
+            result += QString::fromUtf8("...>");
+            level--;
+        }
+        else if (currentChar == QChar(stopBracket) && level > 1) {
+            level--;
+        }
+        else if (level == 0) {
+            result += currentChar;
+        }
+    }
+    return result;
+}
--- a/src/analyze/gui/tests/CMakeLists.txt
+++ b/src/analyze/gui/tests/CMakeLists.txt
@@ -1,0 +1,21 @@
+project(test_middle_elide LANGUAGES CXX)
+
+find_package(QT NAMES Qt5 Qt6 COMPONENTS Test REQUIRED)
+find_package(Qt${QT_VERSION_MAJOR} COMPONENTS Test REQUIRED)
+
+set(CMAKE_INCLUDE_CURRENT_DIR ON)
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+enable_testing()
+
+add_executable(test_middle_elide
+    tst_middle_elide.cpp
+)
+
+add_test(NAME test_middle_elide COMMAND test_middle_elide)
+
+target_link_libraries(test_middle_elide PRIVATE
+                      heaptrack_gui_private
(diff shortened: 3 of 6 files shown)


khReichel:
@milianw 
I have implemented the patch for omitting the template parameters from the flame view (but not from the tooltips, where the information is really necessary).
If you would like to, please add my contribution.
If there is something you would like to have changed, please let me know

src/analyze/gui/middleelide.cpp
```diff
@@ -0,0 +1,42 @@
...
+{
+
+}
+
+QString MiddleElide::elideAngleBracket(const QString& s)
+{
+   return substituteAngleBrackets(s);
+}
+
+QString MiddleElide::substituteAngleBrackets(const QString& s)
```
milianw, line 14: this should just be a free function within `analyze/gui/util.h` - there's no need to put this into a separate file and into a class namespace

src/analyze/gui/middleelide.cpp
```diff
@@ -0,0 +1,42 @@
...
+{
+    static QChar startBracket = QChar(u'<');
```
milianw, line 16: no need to make this static, and use `QLatin1Char` please:

```
const auto startBracket = QLatin1Char('<');
```

src/analyze/gui/middleelide.cpp
```diff
@@ -0,0 +1,42 @@
...
+    static QChar stopBracket  = QChar(u'>');
+
+    int level = 0;
+    QString result;
```
milianw, line 20: `result.reserve(s.size())`

src/analyze/gui/middleelide.cpp
```diff
@@ -0,0 +1,42 @@
...
+    for (int i=0, n = s.length(); i < n; i++)  {
```
milianw, line 21: `for (auto currentChar : s)`

src/analyze/gui/middleelide.cpp
```diff
@@ -0,0 +1,42 @@
...
+        QChar currentChar = s[i];
+        if (currentChar == QChar(startBracket) && level == 0) {
```
milianw, line 23: simplify:

```
if (currentChar == startBracket) {
    if (level == 0) {
        result += startBracket;
    }
    ++level;
} else if (currentChar == stopBracket) {
    if (level == 1) {
        result += stopBracket;
    }
    --level;
} else if (level == 0) {
    result += currentChar;
}
```

src/analyze/gui/middleelide.cpp
```diff
@@ -0,0 +1,42 @@
...
+            result += QChar(startBracket);
+            level++;
+        }
+        else if (currentChar == QChar(startBracket) && level > 0) {
+            level++;
+        }
+        else if (currentChar == QChar(stopBracket) && level == 1) {
```
milianw, line 30: dito

src/analyze/gui/tests/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+project(test_middle_elide LANGUAGES CXX)
```
milianw, line 1: remove this, instead add the test to `tests/auto/CMakeLists.txt`

src/analyze/gui/tests/tst_middle_elide.cpp
```diff
@@ -0,0 +1,64 @@
+#include <QtTest>
+#include <QString>
+#include "analyze/gui/middleelide.h"
+
+// add necessary includes here
```
milianw, line 5: remove

src/analyze/gui/tests/tst_middle_elide.cpp
```diff
@@ -0,0 +1,64 @@
+#include <QtTest>
```
milianw, line 1: missing license header

src/analyze/gui/tests/tst_middle_elide.cpp
```diff
@@ -0,0 +1,64 @@
...
+class test_initilaize : public QObject
+{
+    Q_OBJECT
+
+public:
+    test_initilaize() = default;
+    ~test_initilaize() = default;
+
+private slots:
+    void test_simple_case();
```
milianw, line 25: use data driven tests instead (`QTest::add{Column,Row}` & friends)

src/analyze/gui/tests/tst_middle_elide.cpp
```diff
@@ -0,0 +1,64 @@
...
+const char* nested_brackets = "std::vector<test type <int> in bracket> MainWindow::onLoadingFinish(unsigned int&)";
+const char* nested_brackets_fixed = "std::vector<...> MainWindow::onLoadingFinish(unsigned int&)";
+
+
+class test_initilaize : public QObject
+{
+    Q_OBJECT
+
+public:
+    test_initilaize() = default;
```
milianw, line 21: not needed, remove

src/analyze/gui/flamegraph.cpp
```diff
@@ -192,10 +193,12 @@ void FrameGraphicsItem::paint(QPainter* painter, const QStyleOptionGraphicsItem*
         Q_UNREACHABLE();
     }();
 
+    QString middleElidedLabel = MiddleElide::elideAngleBracket(label);
```
milianw, line 196: this functionality needs to be optional, and should be off-by-default. please add a setting and a gui checkbox to enable it

milianw (changes requested):
hey, thanks for your contribution. there are some ways this could be simplified, but otherwise it's already looking good to me.

src/analyze/gui/middleelide.cpp
khReichel: I was not sure if util.h is the right place for the method.
But it is ok for me

src/analyze/gui/tests/tst_middle_elide.cpp
khReichel: I am familiar with the data-driven tests of the QtTest framework.
However, I wanted to name the individual tests explicitly.
For me, it is much clearer to have individual tests that precisely describe the individual test case.
(e.g. test_single_bracket(), test_multiple_brackets()).

This approach (originally from TDD) describes the tests much more clearly than a data-driven approach with only one method.
What would still be possible in any case is a summary of the redundant code of the individual test methods.

src/analyze/gui/flamegraph.cpp
khReichel: Ok, got it ;-)

khReichel: @milianw Is there some persistancy functionality available in heaptrac.
Otherwise I will use QSettings for storing that elide functionality

milianw: Yes, see `MainWindow::m_config`

src/analyze/gui/tests/tst_middle_elide.cpp
milianw: I don't follow, the rows get a label which would have the same explanation, no? i.e. to me, the following has the same level of detail:

```
testElideTemplates:single_bracket

testElideTemplates_singleBracket
```

milianw:
ping? any progress on this front?

PR #35, opened 2021-08-30
CMake build fixes

jtanx:
* Check if boost iostreams has zstd and use it if available. Prevents conflicting definitions when both are present.
* Rename FindZstd.cmake to FindZSTD.cmake to match the casing of the variables, silences CMake warning.
* Move -Wall from add_compile_definitions to add_compile_options, fixes -Wall from being added to windres.


milianw:
please split this into multiple commits, with one problem solved per patch

jtanx:
Done

milianw:
thank you!

PR #36, opened 2021-12-02
analyze CMakeLists.txt - Provide feedback as to why the gui is not be…

cmorganBE:
…ing built

It took several minutes to figure out what was going on with the gui build. I wanted to find a way to keep the add_subdirectory() outside of the nested conditional but putting it inside lets us provide pretty helpful status messages to the user as to why the gui isn't being built.

The only other thing I stumbled upon was that ECM (extra-cmake-modules) overlaps in name with gmp-ecm, and libecm, which appear to be entirely unrelated. That took a few minutes to figure out and was mostly luck.


milianw:
I don't quite follow, what was the issue you are trying to solve? Can you show me the cmake output before this patch vs. the one after this patch to make the effect more clear?

Generally, cmake already shows messages when an optional or required dependency isn't found, are we maybe missing the explanation as to what that means (e.g. without KF5 no UI can be build?)

cmorganBE:
@milianw let me know if you want me to get the before/after output from the cmake configure step with/without a few of the dependencies.

milianw:
I'll instead just set the purpose for ECM. That gives us:

```
-- The following RECOMMENDED packages have not been found:

 * ECM (required version >= 1.0.0)
   KDE's extra-cmake-modules, required for the heaptrack_gui executable.
```

See also https://github.com/KDE/heaptrack/commit/1a6d63a7b41b9bb8c94fc9a604870b1249f58142

cmorganBE:
@milianw did you want me to make that change or?

Also, does that, when ECM is installed, provide guidance that QT5 and KF5 are required for gui? (As those would be the next things in the build dependency tree required for gui)

milianw:
I made a change already, please see how it works. Then feel free to close this task here

milianw:
Any news? I'm assuming this is resolved now, if now please reopen

cmorganBE:
This should help yep.

PR #38, opened 2022-02-14
Pass client exit code in heaptrack.sh

oleg-liatte:
Needed to detect client failures during heap-tracking.

Note that `set -e` is not applicable here as we still have to `wait $debuggee` even on client failure


milianw:
thank you, merged!

PR #39, opened 2022-02-22
Add support for the mimalloc allocator

MasterDuke17:
This way allocations/free done by it are tracked regardless of whether
it's used via a dynamic override (e.g., preload) or directly via api.

### DIFF
--- a/src/track/heaptrack_inject.cpp
+++ b/src/track/heaptrack_inject.cpp
@@ -53,6 +53,24 @@
 #endif
 #endif
 
+// NOTE: adding noexcept to C functions is a hard error in clang++
+//       (but not even a warning in GCC, even with -Wall)
+#if defined(__GNUC__) && !defined(__clang__)
+#define LIBC_FUN_ATTRS noexcept
+#else
+#define LIBC_FUN_ATTRS
+#endif
+
+extern "C" {
+
+// Foward declare mimalloc (https://github.com/microsoft/mimalloc) functions so we don't need to include its .h.
+void* mi_malloc(size_t size) LIBC_FUN_ATTRS;
+void* mi_calloc(size_t count, size_t size) LIBC_FUN_ATTRS;
+void* mi_realloc(void* p, size_t newsize) LIBC_FUN_ATTRS;
+void  mi_free(void* p) LIBC_FUN_ATTRS;
+
+}
+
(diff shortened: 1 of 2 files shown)


milianw (changes requested):
hey there!

this looks good but you are missing changes to heaptrack_inject which uses a slightly different mechanism and needs to be expanded for these functions too please

MasterDuke17:
> this looks good but you are missing changes to heaptrack_inject which uses a slightly different mechanism and needs to be expanded for these functions too please

I'm not sure how to test the heaptrack_inject mechanism, but this seems likely to be the correct change (i.e., it compiles, ship it!).

MasterDuke17:
Easy enough, and yep, results looked good.

milianw:
I just tried it on an app that's not using mimalloc and that then fails b/c we fail to find the original mimalloc entry points. the code needs to be changed to take that into account - i.e. this shouldn't be a fatal error for the mimalloc functions:

```
Could not find original function mi_malloc
```

MasterDuke17:
My bad. I'm not sure how you want this to be handled, but
```patch
diff --git src/track/heaptrack_preload.cpp src/track/heaptrack_preload.cpp
index 4147f15..b5d4b5f 100644
--- src/track/heaptrack_preload.cpp
+++ src/track/heaptrack_preload.cpp
@@ -54,7 +54,7 @@ struct hook
     void init() noexcept
     {
         auto ret = dlsym(RTLD_NEXT, Base::identifier);
-        if (!ret) {
+        if (!ret && strstr(Base::identifier, "mi_") != Base::identifier) {
             fprintf(stderr, "Could not find original function %s\n", Base::identifier);
             abort();
         }
```
seems to work for me (i.e., everything looks the same for mimalloc-using applications, and no errors+normal-looking results for non-mimalloc-using applications).

MasterDuke17:
@milianw should I push the patch I showed? Or do you have a different way in mind?

milianw:
yes, that would be acceptable. Any alternative approach I could then add after the fact. I'll handle this now and merge the code. Note that this will happen outside the github ui (b/c heaptrack is actually only mirrored here, the real upstream is over at invent.kde.org)

milianw:
Done now, many thanks again for your contribution

PR #40, opened 2022-04-08
handle calling heaptrack_stop() without debuginfo

llunak:
If the heaptrack inject binary is stripped (as may easily be the case e.g. with distribution packages), gdb prints out "'heaptrack_stop' has unknown return type; cast the call to its declared return type". Add an explicit cast, like all other gdb calls already use.


milianw:
Thank you!

PR #42, opened 2022-10-13
Does this tool check virtual memory or physical memory ?

DanielKui:
as titled


milianw:
wtf? don't abuse pull requests to ask questions, use the mailing list

PR #1, opened 2015-12-02
Make number of reported peaks configurable

dajva:
I find the hard coded peak limit of 10 a bit restrictive. Hope you are interested in this patch.

PR #43, opened 2022-10-13
1.3

DanielKui:
kb@debian11:~/QuoteST$ /home/kb/heaptrack-1.3.0/build/bin/heaptrack ./QuoteSvr
heaptrack output will be written to "/home/kb/QuoteST/heaptrack.QuoteSvr.3353971.gz"
/home/kb/heaptrack-1.3.0/build/lib/heaptrack/libheaptrack_preload.so
starting application, this might take some time...
./QuoteSvr: symbol lookup error: /home/kb/heaptrack-1.3.0/build/lib/heaptrack/libheaptrack_preload.so: undefined symbol: _ULx86_64_set_cache_size

When I run the above command, it prompts "symbol lookup error", then the program have been blocking here. no any result display.

My env:
Linux debian11 5.10.0-10-amd64 #1 SMP Debian 5.10.84-1 (2021-12-08) x86_64 GNU/Linux

### DIFF
--- a/src/analyze/gui/org.kde.heaptrack.appdata.xml
+++ b/src/analyze/gui/org.kde.heaptrack.appdata.xml
@@ -57,7 +57,7 @@
       Heaptrack traces all memory allocations and annotates these events with stack traces. Dedicated analysis tools then allow you to interpret the heap memory profile to:
     </p>
     <p xml:lang="ca">El «heaptrack» rastreja totes les assignacions de memòria i anota aquests esdeveniments amb seguiments de pila. Les eines d'anàlisi dedicades que hi ha a continuació, us permetran interpretar el perfil de memòria en monticles:</p>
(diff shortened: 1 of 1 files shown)


milianw:
use the official bug tracker to report issues, instead of abusing pull requests for this purpose

generally it seems like you have a setup issue with libunwind - it doesn't come with `unw_set_cache_size` which we detecth in `FindLibunwind.cmake` - try to nuke your build dir and start from scratch

dufferzafar:
My application failed with the exact same error. I wasn't using a self-compiled Heaptrack, but a downloaded v1.3.0 from KDE's site.

Heaptrack was working fine with my app before, what changed was that I started linking my app with gperftools profiler & libunwind. Trying to remove that as a dep now.

dufferzafar:
Yes, that was the issue. I removed the gperftools profiler & it's dependency libunwind (1.2.1) from my app and it started working with heaptrack again.

PR #46, opened 2023-04-18
doc: update helptext with record_only option

rwooster:
The --record_only option is useful to know about, but it is not currently documented in the --help output of the heaptrack executable. Add documentation for this option.


milianw:
thank you, I pushed this now. note that github is a read-only mirror so I cherry picked and will close the request here

cheers, thanks for the contribution

PR #47, opened 2023-06-12
bugfix: wrapper script drops replacements for output parameter

lukasz-m-maciejewski:


milianw:
thanks, pushed. note that github is read-only so I manually cherry picked and pushed this. as such, I'm closing this now

cheers

PR #49, opened 2023-11-15
Use boost::hash_value for IndexedAllocationInfo

glandium:
The current implementation doesn't really work well with 64-bits values, and can lead to the robin-set using excessive amounts of memory because of the number of collisions that can happen.

See https://bugs.kde.org/show_bug.cgi?id=477018


milianw:
Sorry, I didn't see this MR and fixed it myself just now...

PR #50, opened 2024-01-14
[Bug 379660] Collapse template arguments in flamegraph using checkbox

aravindev:
This PR is a duplicate of https://github.com/KDE/heaptrack/pull/34 but addresses all the comments since it has been dormant for a while.

I have created a checkbox that can be used to toggle template elision such that the template arguments can optionally be collapsed. This feature is disabled by default.

@milianw Since you were the original reviewer, could you take a look?


aravindev:
@SamSaffron @iamsergio @krf @milianw Could one of you assign a reviewer? If this is no longer of interest, feel free to close this PR :smile:

milianw:
Sorry for the delay, I'll have a look now

milianw (changes requested):
So, this code is potentially a first step but I think it should be closer to the approach we use in hotspot - see e.g.:

https://github.com/KDAB/hotspot/blob/master/src/util.cpp#L135

generally, we would collapse these _everywhere_ - not just in the flamegraph. the same logic should be used in the tables, and we don't want to do the collapse all the time for view purposes only

aravindev:
> So, this code is potentially a first step but I think it should be closer to the approach we use in hotspot - see e.g.:
> 
> https://github.com/KDAB/hotspot/blob/master/src/util.cpp#L135
> 
> generally, we would collapse these _everywhere_ - not just in the flamegraph. the same logic should be used in the tables, and we don't want to do the collapse all the time for view purposes only

Thanks for the pointers. I will update the PR and ping you again!

ImAwsumm (approved):

PR #51, opened 2024-02-28
use user ID instead of name

JohnFinn:
Discovered that when using heaptrack across docker containers with shared PID namespaces the following error is produced:
```
$ heaptrack -p <some-PID>

chown: invalid user: 'UNKNOWN'
```
This simple change enables usage of heaptrack in this scenario


milianw:
merged now into the upstream repo over at https://invent.kde.org/sdk/heaptrack/

PR #54, opened 2025-01-11
Check stringId.index before accessing the string

tedyu:
This PR adds checking for the case where `stringId.index` is less than 1.


tedyu:
cc @milianw

milianw:
and when does this happen? what is motivating this change?

tedyu:
I use `heaptrack` to detect memory leaks.

When I was going over the code, I found this place where check for invalid index was missing.

I am yet to reproduce the case where exception happens.

milianw:
the index is never set to a negative number anywhere (isn't it even an unsigned value?). and we already check for null. as far as I can tell this patch is bogus.

please provide a backtrace or better explanation on how to reproduce and reopen, for now I'm going to close this

PR #56, opened 2025-01-31
Add link to heaptrack-trim

untitaker:


milianw (commented):
while I'm not opposed to reference this, why is this not included in the main repo? i.e. why don't you upstream your code into this repo here? or is that not an option due to cargo/rust?

untitaker:
it's possible and up to you, I don't really mind either way. if it's in this repo I guess bug reports will go to the kde's gitlab issuetracker.

milianw:
Personally I think this functionality should ideally come directly with the code base, but then be written in C++ and reuse the existing parser logic. So for now let's reference your external repo and I invite you or anyone else to implement this directly here.

PR #57, opened 2025-03-17
Check to resolve files using their full path against --extra-paths

Ipiano:
The paths passed with --extra-paths are treated as the full path to a directory containing debug symbols; but for very large projects which deploy executables and shared objects together, this can be cumbersome to hunt down and list all those paths.

If such a project is deployed from a local staging directory (e.g. by running DESTDIR=./staging make install), this change will cause heaptrack to find debug symbols if that staging directory is given as an extra path.

Example:
Given the following directory tree

staging
  opt
    myproject
      bin
        myexecutable
        .debug
      lib
        somelib.so
        otherlib.so
        .debug
      qml
        qmlplugin.so
        .debug

Assuming that the files are deployed with something like
    rsync staging/opt remotehost:/

It would be necessary to interpret a raw heaptrack report using 3 extra paths, one for each of the .debug folders.

With this change, it works to use just one extra path pointing to the staging directory, because the relative paths to the files on the device are same as they are relative to the staging directory.


src/interpret/heaptrack_interpret.cpp
```diff
@@ -304,7 +304,36 @@ struct AccumulatedTraceData
         dwfl_end(m_dwfl);
     }
 
-    /// find a file in the sysroot or extra path
+    /// find a file in the sysroot or extra path and add it to the cache
+    const std::string& resolveAndCacheFile(const std::string& fileName)
+    {
+        if (!m_extraPaths.empty()) {
+            // look for the filename, checking both if it's there using the defined path
```
milianw, line 311: imo, we should do first one loop that finds the files using one approach (e.g. the full path, like you propose), and then do a second loop as a fallback with the other approach

the mixed approach of your code might behave in funky ways when there are multiple extra paths, no?

src/interpret/heaptrack_interpret.cpp
```diff
@@ -304,7 +304,36 @@ struct AccumulatedTraceData
...
+            const auto baseName = std::filesystem::path(fileName).filename().string();
+            for (const auto& extraPath : m_extraPaths) {
+                auto fileInExtraPath = extraPath + '/' + baseName;
+                if (fileExists(fileInExtraPath))
+                    return m_resolvedFiles.insert(m_resolvedFiles.end(), {fileName, std::move(fileInExtraPath)})
+                        ->second;
+
+                fileInExtraPath = extraPath + '/' + fileName;
+                if (fileExists(fileInExtraPath))
+                    return m_resolvedFiles.insert(m_resolvedFiles.end(), {fileName, std::move(fileInExtraPath)})
```
milianw, line 327: DRY: make this a lambda and use it in both places

milianw (changes requested):
sounds like a good idea to me in general!

src/interpret/heaptrack_interpret.cpp
Ipiano: Resolved this and the other comment about search order together - let me know if that's a step too far; I can back off it it's too complicated of a solution :smile:

src/interpret/heaptrack_interpret.cpp
milianw: the way your new version (and my adapted one) handle it is fine - at least we'll always prefer one location over the other now, independent of how we traverse the folder structure.

milianw:
Also: please squash your commits

ImAwsumm (approved):

src/interpret/heaptrack_interpret.cpp
```diff
@@ -304,7 +318,39 @@ struct AccumulatedTraceData
...
 
-    /// find a file in the sysroot or extra path
+    /// find a file in the sysroot or extra path and add it to the cache
+    const std::string& resolveAndCacheFile(const std::string& fileName)
+    {
+        const auto baseName = std::filesystem::path(fileName).filename().string();
+
+        const auto tryResolvePath =
+            [this, fileName](const std::string& basePath,
+                             const std::string& subPath) -> std::optional<std::reference_wrapper<const std::string>> {
```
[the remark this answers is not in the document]
Ipiano: Done

Ipiano:
> my code is untested, please double check.

Tested - had to add a `->second` to the final return

milianw (approved):
lgtm now, thanks. I'll merge it manually through the KDE git, note that github here is just a mirror

PR #59, opened 2025-12-11
Fix usage of nonstandard uint type

jtanx:
This type doesn't exist on windows; e.g.

```
/heaptrack/src/util/pointermap.h:63:46: error: 'uint' does not nam
e a type; did you mean 'int'?
   63 |         allocationIndex->index = static_cast<uint>(set.size());
      |                                              ^~~~
      |                                              int
```

### DIFF
--- a/src/analyze/gui/locationdata.h
+++ b/src/analyze/gui/locationdata.h
@@ -75,7 +75,7 @@ struct hash<Symbol>
 };
 }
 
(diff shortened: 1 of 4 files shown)


milianw:
should be fixed by 8dacbc40fd4e8d36eb0ca8663c90b0d6bb661568

PR #60, opened 2026-07-02
perf: add thread-local TraceCache to skip redundant TraceTree lookups

azurezene:
In allocation-heavy workloads, every malloc/free must walk the shared TraceTree trie to resolve its backtrace. This lookup holds a lock and consumes significant critical-section time, causing contention under multi-threaded workloads.

However, high-frequency allocation paths tend to be highly repetitive — the same call site allocating in a loop, the same function calling malloc repeatedly. The backtrace is identical every time, yet we pay the full cost of walking the trie on each call.

This PR inserts a thread-local 8-entry LRU cache in front of TraceTree::index() to skip redundant trie traversals. A cache hit avoids both the trie walk and the associated lock contention entirely.

Implementation:

Key: hash of the backtrace (Murmur-style via 0xbb67ae8584caa73b), with memcmp verification of the first 4 IP frames to eliminate false positives from hash collisions.
Eviction: decay-based rather than exact LRU — each miss decays the tail entry's hit counter by 25%; entries drop below the replacement threshold after ~4 consecutive misses. Hot entries resist eviction naturally.
Cache-line aware layout: key[] in the first cache line (read on every lookup), hitCount/index in the second (touched only on hit), IP storage in colder lines — minimizing cache misses on the fast path.
Lock-free: thread_local storage means zero contention, unlike any shared cache structure.

Performance：

This is the flame graph proportion for /heaptrack/tests/manual/threaded.cpp. It can be clearly seen that the proportion of the TraceTree::index function within the heaptrack_malloc function has decreased from 31.1% to 0.4%.

without cache:
图片

with cache:
图片

I found This idea was originally suggested in [#13](https://github.com/KDE/heaptrack/pull/13) but never landed

### DIFF
--- a/src/track/libheaptrack.cpp
+++ b/src/track/libheaptrack.cpp
@@ -39,6 +39,7 @@
 #include <string>
 #include <thread>
 
+#include "trace_cache.h"
 #include "tracetree.h"
 #include "util/config.h"
 #include "util/libunwind_config.h"
@@ -112,6 +113,13 @@ struct RecursionGuard
 };
 
 thread_local bool RecursionGuard::isActive = false;
+
+/**
+ * Per-thread cache of backtrace → TraceTree index mappings.
+ * Avoids the shared TraceTree trie walk for repeated backtraces
+ * without any lock contention.
+ */
+static thread_local TraceCache t_traceCache;
 
 enum DebugVerbosity
 {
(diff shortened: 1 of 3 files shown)


src/track/trace.h
```diff
@@ -77,7 +85,16 @@ struct Trace
 private:
     int m_size = 0;
     int m_skip = 0;
+    size_t m_key = 0;
     ip_t m_data[MAX_SIZE];
+
+    size_t getHashKey() const
+    {
+        size_t h = (size_t)m_size * 0xbb67ae8584caa73bULL;
```
milianw, line 93: please modernize this function, no c-style casts but static cast instead, also below. and document that this is the murmur hash you mention in the PR

src/track/trace.h
```diff
@@ -42,6 +43,11 @@ struct Trace
         return m_size;
     }
 
+    size_t hashkey() const
```
milianw, line 46: please rename to `hashKey`

src/track/trace.h
```diff
@@ -42,6 +43,11 @@ struct Trace
...
+    {
+        return m_key;
```
milianw, line 48: this should also be called `m_hashKey`

src/track/trace.h
```diff
@@ -77,7 +85,16 @@ struct Trace
 private:
     int m_size = 0;
     int m_skip = 0;
+    size_t m_key = 0;
     ip_t m_data[MAX_SIZE];
+
+    size_t getHashKey() const
```
milianw, line 91: please rename to `computeHashKey()`

src/track/libheaptrack.cpp
```diff
@@ -526,15 +534,21 @@ class HeapTrack
         }
         updateModuleCache();
 
-        const auto index = s_data->traceTree.index(trace, [](uintptr_t ip, uint32_t index) {
-            // decrement addresses by one - otherwise we misattribute the cost to the wrong instruction
-            // for some reason, it seems like we always get the instruction _after_ the one we are interested in
-            // see also: https://github.com/libunwind/libunwind/issues/287
-            // and https://bugs.kde.org/show_bug.cgi?id=439897
-            --ip;
+        uint32_t index = t_traceCache.lookup(trace);
```
milianw, line 537: I would subjectively prefer if the API would change and the hashing gets moved into the new trace cache. I.e. the Trace itself doesn't need to be changed (I think), so instead you'd do something like:

```
const auto [index, hash] = t_traceCache.lookup(trace);
...
if (index) {
    t_traceCache.store(trace, index, hash);
}
```

but again: very subjective. if you prefer your way then that's fine with me I guess.

milianw (changes requested):
sorry for the delay - some small nits, the idea is great - but is it sound?

most notably, aren't we suffering from potential false-positive lookups now? when we get a hash collision, then there's no guarantee from the cache to prevent that, no? there is only a heuristic or "best effort", or am I missing something? not saying this is a no-go, I'd like to better understand why you are saying this is sound and safe.

generally, can you please also add some tests for the new trace cache, to ensure this works as intended?

